PR #4, opened 2016-07-22
Adding an explicit content type to the github asset uploader

jer:
- The `file` command doesn't exist in the google/golang box, which
  causes a bug when running github-upload-asset


Filbird:
👍

PR #5, opened 2016-07-25
Add shell script to auto increment version number

Filbird:
This script will increment the patch number with every Wercker build.
VERSION file gets updated with new value.

### DIFF
--- a/wercker.yml
+++ b/wercker.yml
@@ -13,11 +13,17 @@ build:
 deploy:
   steps:
     - script:
-        name: get version from VERSION file
-        code: export VERSION=$(cat VERSION)
+        name: get highest version from tags
+        code: export VERSION=$(sh incrementVersion.sh)
+    - script:
+        name: add new annotated tag
+        code: git tag -a $VERSION -m "Tagging release version $VERSION"
+    - script:
+        name: push newly added tag
+        code: git push origin $VERSION    
     - github-create-release:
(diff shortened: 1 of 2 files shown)


stjohnjohnson:
Shouldn't you be using tags?  If not, you need to check this file in each time.  @jer

jer:
Yep, good start but we need to at least get all git tags and sort them numerically. Then we need to bump the largest version, tag the repo, push the tags.

wercker.yml
```diff
@@ -13,11 +13,17 @@ build:
 deploy:
   steps:
     - script:
-        name: get version from VERSION file
-        code: export VERSION=$(cat VERSION)
+        name: get highest version from tags
+        code: export VERSION=$(sh incrementVersion.sh)
+    - script:
+        name: add new annotated tag
+        code: git tag -a $VERSION -m "Auto-generated tag message"
```
Filbird: What annotated message should the tag have?

Filbird:
Shell script now gets the version from the highest git tag value rather than a local file.
Added steps in the wercker.yml to generate a new tag and push it.

wercker.yml
jer: How about "Tagging release version x.y.z"?

Filbird: sure! Will change

jer:
👍

PR #10, opened 2016-07-27
Add 'install-packages' step to install openssh-client.

Filbird:
This should enable the "add-to-known-hosts" step to run


stjohnjohnson:
Wow, this is complicated. :( 👍

jer:
Ooof. Maybe there's a better image
👍

Issue #13, opened 2016-07-27
Replace Git Clone and Parse operation


stjohnjohnson:
What you need to do:

1. Clone the git repository and branch to a directory named `$org/$repo` (https://github.com/screwdriver-cd/job-tools/blob/master/launch#L18-L19)
2. Setup artifacts_dir (create and export): https://github.com/screwdriver-cd/job-tools/blob/master/launch#L22-L25
3. Post the `screwdriver.yaml` to `$api_url/validator`
4. Export `jobs.$jobName.environment` to `artifacts_dir/environment.json`
5. Export `jobs.$jobName.steps` to `artifacts_dir/execute.json`
6. Run the executor: https://github.com/screwdriver-cd/job-tools/blob/master/launch#L33

Input to script is still:
 - $org
 - $repo
 - $branch
 - $jobName

stjohnjohnson:
While we're there, we should add `--depth=50 --quiet --progress` to the clone operation

stjohnjohnson:
Done!

PR #16, opened 2016-07-29
Add clone and artifacts in Go

Filbird:
Functionality for items 1.) and 2.) in #13 

TODO: 
-Look over golang specific testing patterns/packages that provide better output and display code coverage
-Integrate remaining items from #13 
-Refactor to reduce redundancy of error checking?
-Uncomment "git config user.name" and "git config user.email"

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -1,41 +1,55 @@
 package main
 
-import (
+import(
 	"fmt"
 	"os"
-	"os/exec"
 )
 
-type shellOutInterface interface {
-	run(string, ...string) ([]byte, error)
-}
+var SCRIPT_DIR string
+var GIT_ORG string
+var GIT_REPO string
+var GIT_PATH string
+var GIT_BRANCH string
+var JOB_NAME string
 
-var shellOut shellOutInterface
 
-type shellOutImpl struct{}
-
-func init() {
-	shellOut = shellOutImpl{}
-}
-
-func (s shellOutImpl) run(cmd string, args ...string) (out []byte, err error) {
-	return exec.Command(cmd, args...).CombinedOutput()
-}
-
-func runLauncher(args ...string) (out []byte, err error) {
-	cmdName := "/opt/screwdriver/launch.sh"
-	return shellOut.run(cmdName, args...)
-}
-
-func main() {
-	var out []byte
-	var err error
-
-	fmt.Println("Launching the launcher...")
-	if out, err = runLauncher(os.Args[1:]...); err != nil {
-		fmt.Fprintln(os.Stderr, "There was an error shelling out to the launcher: ", err)
+func main(){
+	//Set up arguments
+	if len(os.Args) < 3 {
+		fmt.Fprintln(os.Stderr, "Missing arguments: {GIT_ORG, GIT_REPO, GIT_BRANCH, JOB_NAME}")
 		os.Exit(1)
 	}
-	fmt.Println(string(out))
-	fmt.Println("Launched successfully.")
+	GIT_ORG = os.Args[1]
+	GIT_REPO = os.Args[2]
+	GIT_PATH = "github.com/" + GIT_ORG + "/" + GIT_REPO
+	GIT_BRANCH = "master"
+	if len(os.Args) >= 4 {
+		GIT_BRANCH = os.Args[3]
+	}
+	JOB_NAME = "main"
+	if len(os.Args) == 5 {
+		JOB_NAME = os.Args[4]
+	}
+
+	//Shell commands to execute in order
+	err := Clone(GIT_PATH, GIT_BRANCH)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "There was an error cloning the git repository: ", err)
+		os.Exit(1)
+	}
+
+	_, err = FetchAndMergePR(GIT_BRANCH, JOB_NAME)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "There was a fetching or merging the pull request ", err)
+		os.Exit(1)
+	}
+
+	err = ArtifactsDirSetUp()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "There was an error setting up the artifacts directory: ", err)
+		os.Exit(1)
+	}
+
+
+	//Call remaining shell commands here:
 }
--- a/launch_test.go
+++ b/launch_test.go
@@ -1,42 +1,156 @@
 package main
 
-import (
-	"fmt"
+import(
+	"os"
+	"os/exec"
 	"strings"
 	"testing"
 )
+var baseDir string
 
-var shellOutSetup map[string]string
-
-type shellOutMock struct{}
-
-func (s shellOutMock) run(cmd string, args ...string) (out []byte, err error) {
-	cmdslice := []string{cmd}
-	cmdslice = append(cmdslice, args...)
-
-	cmd = strings.Join(cmdslice, " ")
-	output := shellOutSetup[cmd]
-	if output == "" {
-		return nil, fmt.Errorf("Error: %s not mocked", cmd)
-	}
-
-	return []byte(output), nil
+// Create a test directory and change into it
+func setUp() {
+	os.Mkdir("testDir", 0777)
+	os.Chdir("testDir")
 }
 
-func Test(t *testing.T) {
-	shellOut = shellOutMock{}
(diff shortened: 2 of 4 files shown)


shell/shell.go
```diff
@@ -0,0 +1,106 @@
...
+	matched, err := regexp.MatchString(reg, job)
+	if err != nil {
+		return "", err
+	}
+
+	if matched == true {
+		PR_NUMBER = strings.Split(job, "-")[1]
+
+		// ShellCommand("git", "config", "user.name", "sd-buildbot")
+		// ShellCommand("git", "config", "user,email", "devnull@screwdriver.cd")
```
stjohnjohnson: `user.email`

Filbird:
Trying to figure out why "go test" is failing in the wercker pipeline. Seems to be working locally. Is it possible that the PR has to go through first in order for this step to find the package?

launch.go
```diff
@@ -1,41 +1,55 @@
...
+import(
 	"fmt"
 	"os"
-	"os/exec"
 )
 
-type shellOutInterface interface {
-	run(string, ...string) ([]byte, error)
-}
+var SCRIPT_DIR string
```
jer: Are these global variables necessary?

launch_test.go
```diff
@@ -1,42 +1,156 @@
...
-	cmdslice := []string{cmd}
-	cmdslice = append(cmdslice, args...)
+	wdTest, _ := ShellCommand("pwd")
 
-	cmd = strings.Join(cmdslice, " ")
-	output := shellOutSetup[cmd]
-	if output == "" {
-		return nil, fmt.Errorf("Error: %s not mocked", cmd)
+	// Working directories should be equal
+	if strings.Compare(wd, wdTest) != 0 {
```
jer: Can't you do wd != wdTest instead?

launch.go
```diff
@@ -1,41 +1,55 @@
...
-type shellOutImpl struct{}
-
-func init() {
-	shellOut = shellOutImpl{}
-}
-
-func (s shellOutImpl) run(cmd string, args ...string) (out []byte, err error) {
-	return exec.Command(cmd, args...).CombinedOutput()
-}
+func main(){
```
jer: Can we split this up into more functions to make the functions more testable?

launch_test.go
```diff
@@ -1,42 +1,156 @@
...
+//Test that Clone will return an error if GIT_PATH or GIT_BRANCH is incorrect
+func TestCloneBadInput(t *testing.T) {
+	setUp()
+
+	// Arguments
+	GIT_PATH := "badPath"
+	GIT_BRANCH := "badBranch"
+
+	// Error message should not be nil
+	err := Clone(GIT_PATH, GIT_BRANCH)
```
jer: We should mock out what happens during clone rather than actually cloning for every test. We know git works, we just want to know that we're calling git correctly at this point.

We need functional tests at some point that exercise everything end-to-end.

launch_test.go
```diff
@@ -1,42 +1,156 @@
...
+	if err == nil {
+		t.Logf("Expected err to not be: nil")
+		t.Fail()
 	}
+
+	cleanUp()
 }
+
+// // Test that FetchAndMergePR will return without errors and return the correct pull request number
```
jer: Do we need all of these commented-out lines?

launch_test.go
```diff
@@ -1,42 +1,156 @@
...
+func TestArtifactsDirSetUp(t *testing.T) {
+	setUp()
+
+	// Call Clone
+	GIT_PATH := "github.com/Filbird/pr"
+	GIT_BRANCH := "master"
+	_ = Clone(GIT_PATH, GIT_BRANCH)
+
+	err := ArtifactsDirSetUp()
+	// Error message should be nil
```
jer: This comment doesn't really add value considering that the code below says the same thing

launch_test.go
```diff
@@ -1,42 +1,156 @@
...
+	// Working directory should be updated
+	if strings.Compare(wd, wdTest) != 0 {
+		t.Logf("Expected working directory to equal: %s, but was: %s", wd, wdTest)
+		t.Fail()
 	}
-	if err != nil {
-		errmsg := fmt.Sprintf("Error running command: %s", err)
-		t.Error(errmsg)
+
+	cleanUp()
```
jer: This line will not be reached if line 63 is triggered. Cleanup type things that have to run should be called with defer at the top of the function.

That being said, setup/cleanup across each test indicates a problem with the test in my eyes.

launch_test.go
```diff
@@ -1,42 +1,156 @@
...
+	wd, _ := ShellCommand("pwd")
 
-	shellOutSetup = map[string]string{
-		"/opt/screwdriver/launch.sh abc": "abc",
+	// Arguments
+	GIT_PATH := "github.com/Filbird/pr"
+	GIT_BRANCH := "master"
+
+	err := Clone(GIT_PATH, GIT_BRANCH)
+	// Error message should be nil
```
jer: We know!

launch_test.go
```diff
@@ -1,42 +1,156 @@
...
+		t.Logf("Expected err to be: nil, but was %v", err)
+		cleanUp()
+		t.FailNow()
+	}
+
+	wd, _ := ShellCommand("pwd")
+	artifactsDir := os.Getenv("ARTIFACTS_DIR")
+	artifactsDir = artifactsDir + "\n"
+	// Artifacts directory should be exported
+	if strings.Compare(wd, artifactsDir) != 0 {
```
jer: What exactly is being tested here? Is this saying that when I call ArtifactsDirSetUp(), my current directory should change and ARTIFACTS_DIR gets set in the environment? And those 2 values should be the same? I'm not feeling the chdir aspect of this function.

launch_test.go
```diff
@@ -1,42 +1,156 @@
...
+		t.FailNow()
 	}
 
-	output, err := runLauncher("abc")
-	if string(output) != shellOutSetup["/opt/screwdriver/launch.sh abc"] {
-		errmsg := fmt.Sprintf("Expected %s, got %s", shellOutSetup["echo"], output)
-		t.Error(errmsg)
+	wdTest, _ := ShellCommand("pwd")
+	wd = strings.Trim(wd, "\n")
+	wd = wd + "/" + GIT_PATH + "\n"
```
jer: I think fmt.Srpintf is more idiomatic

launch_test.go
```diff
@@ -1,42 +1,156 @@
...
+	// Working directory should be updated
+	if strings.Compare(wd, wdTest) != 0 {
```
jer: wd != wdTest?

launch_test.go
```diff
@@ -1,42 +1,156 @@
...
+		t.Logf("Expected working directory to equal: %s, but was: %s", wd, wdTest)
```
jer: You can just do t.Failf("Expected working directory to equal: %s, but was: %s", wd, wdTest)

shell.go
```diff
@@ -0,0 +1,108 @@
...
+// specified arguments. It prints the output of the shell command.
+func ShellCommand(cmd string, args ...string) (out string, err error) {
+	var output []byte
+
+	output, err = exec.Command(cmd, args...).CombinedOutput()
+	if err != nil {
+		return "", err
+	}
+
+	return string(output), err
```
jer: Isn't this combined with the err check above 100% identical to a function that looks like this?

```go
func ShellCommand(cmd string, args ...string) (string, error) {
  return exec.Command(cmd, args...).CombinedOutput()
}
```

Except that it casts one of the return values to a string?

jer: Seems like people could just call the exec.Command version and skip the middleman.

jer: The main reason you would wrap the call is if you are using a mocked version in tests. I don't think you're doing that, are you?

shell.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// Clone will clone the specified branch from the repository in path 
+// and switch to its directory.
+func Clone(path string, branch string) (err error) {
```
jer: The name `path` seems misleading here, since it's actually most of a url (without https://)

jer: Perhaps better would be to accept an org and a repo, or perhaps to accept an entire valid url for git checkout, and parse that to calculate a checkout directory.

stjohnjohnson:
Replacing with new PRs

PR #21, opened 2016-08-04
Added handling of non-200 responses from Screwdriver gets

jer:
Also refactored the Authorization header test to de-dupe


stjohnjohnson:
👍 
oh_go

Issue #15, opened 2016-07-27
Support Pull Requests


stjohnjohnson:
If the job name matches `/^PR-\d+/`, then it is a pull request job.  Pull request jobs should:

1. Get the PR id (it's the `\d+`)
1. Clone from the desired branch
1. Fetch `pull/${PR_NUMBER}/head` branch
1. Merge onto the current branch
1. Change the job name to `main`
1. Continue as normal

PR #26, opened 2016-08-04
Add checkIfPR, Fetch, and Merge functions

Filbird:
Adds functionality for checking if job name is a PR, then fetching and merging if so.
Addresses #15. (except updating job name)

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -10,6 +10,24 @@ import (
 
 var execCommand = exec.Command
 
+// command executes the git command
+func command(arguments ...string) error {
+	cmd := execCommand("git", arguments...)
+
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	err := cmd.Start()
+	if err != nil {
+		return fmt.Errorf("starting git command: %v", err)
+	}
+
+	err = cmd.Wait()
+	if err != nil {
+		return fmt.Errorf("running git command: %v", err)
+	}
+	return nil
+}
+
 // Clone clones a git repo into a destination directory
 func Clone(scmURL, destination string) error {
 	log.Printf("Cloning %v", scmURL)
@@ -20,36 +38,33 @@ func Clone(scmURL, destination string) error {
 
 	repo := parts[0]
 	branch := parts[1]
-	cmd := execCommand("git", "clone", "--quiet", "--progress", "--branch", branch, repo, destination)
-
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
-	err := cmd.Start()
-	if err != nil {
-		return fmt.Errorf("starting git clone command: %v", err)
-	}
-
-	err = cmd.Wait()
-	if err != nil {
-		return fmt.Errorf("cloning git repo: %v", err)
-	}
-	return nil
+	return command("clone", "--quiet", "--progress", "--branch", branch, repo, destination)
 }
 
-//SetConfig sets up git configuration
+// SetConfig sets up git configuration
 func SetConfig(setting, name string) error {
-	cmd := execCommand("git", "config", setting, name)
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
+	return command("config", setting, name)
+}
 
-	err := cmd.Start()
+// FetchPR fetches a pull request into a specified branch
+func FetchPR(prNumber string, branch string) error {
+	return command("fetch", "origin", "pull/"+prNumber+"/head:"+branch)
+}
+
+// Merge merges changes on the specified branch
+func Merge(branch string) error {
+	return command("merge", "--no-edit", branch)
+}
+
+// MergePR calls FetchPR and Merge
+func MergePR(prNumber string, branch string) error {
+	err := FetchPR(prNumber, branch)
 	if err != nil {
-		return fmt.Errorf("starting git config command: %v", err)
+		return fmt.Errorf("fetching PR: %v", err)
 	}
-
-	err = cmd.Wait()
+	err = Merge(branch)
 	if err != nil {
-		return fmt.Errorf("setting git config: %v", err)
+		return fmt.Errorf("merging PR: %v", err)
 	}
 	return nil
 }
--- a/git/git.go
+++ b/git/git.go
@@ -9,6 +9,11 @@ import (
 )
 
 var execCommand = exec.Command
+var clone = Clone
+var setConfig = SetConfig
+var mergePR = MergePR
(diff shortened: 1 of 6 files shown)


git/git.go
```diff
@@ -22,3 +23,37 @@ func Clone(repo, destination string) error {
...
+	err = cmd.Wait()
+	if err != nil {
+		return fmt.Errorf("Fetching pull request: %v", err)
+	}
+	return nil
+}
+
+// Merge merges changes on the "_pr" branch
+func Merge() error {
+	cmd := execCommand("git", "merge", "_pr")
```
Filbird: Should we combine Fetch and Merge into a single function? 
I think the only time we merge is right after fetching.
If so would need to alter test code so as to mock 2 shell commands from a single function

launch.go
```diff
@@ -23,7 +24,11 @@ type scmPath struct {
 
 var mkdirAll = os.MkdirAll
 var stat = os.Stat
+var matchString = regexp.MatchString
+var checkIfPr = checkIfPR
 var gitClone = git.Clone
```
Filbird: Okay to mock out internal package function?

launch.go
```diff
@@ -23,7 +24,11 @@ type scmPath struct {
 
 var mkdirAll = os.MkdirAll
 var stat = os.Stat
+var matchString = regexp.MatchString
+var checkIfPr = checkIfPR
```
Filbird: Okay to mock out internal package function?

launch.go
```diff
@@ -88,6 +93,23 @@ func createWorkspace(srcPaths ...string) (Workspace, error) {
...
+	reg := "^PR-[0-9]+$"
+
+	matched, err := matchString(reg, jobName)
+	if err != nil {
+		return prNumber, nil
+	}
+
+	if matched == true {
+		prNumber = strings.Split(jobName, "-")[1]
+	}
```
Filbird: Need to mock out strings.Split?

launch_test.go
```diff
@@ -228,3 +230,51 @@ func TestClone(t *testing.T) {
...
+		t.Errorf("Unexpected error checking pull request")
+	}
+
+	if prNumber != wantPrNumber {
+		t.Errorf("prNumber == %q, want %q", prNumber, wantPrNumber)
+	}
+}
+
+func TestFetchAndMerge(t *testing.T) {
+	testBuildID := "BUILDID"
```
Filbird: Split tests? Feels redundant to do so and doesn't generate further code coverage.

jer:
There's like 100 commits here. Something went wrong in your rebasing.

git/git.go
jer: I think it's fine to have separate functions, but I wouldn't object to a function that combines things like MergePR or something

git/git.go
```diff
@@ -22,3 +23,37 @@ func Clone(repo, destination string) error {
...
 }
+
+// Fetch fetches a pull request into a branch named "_pr"
+func Fetch(prNumber string) error {
+	cmd := execCommand("git", "fetch", "origin", "pull/"+prNumber+"/head:_pr")
+	cmd.Stdout = os.Stdout
+
+	err := cmd.Start()
+	if err != nil {
+		return fmt.Errorf("starting git fet command: %v", err)
```
jer: s/fet/fetch/g

git/git.go
```diff
@@ -22,3 +23,37 @@ func Clone(repo, destination string) error {
...
+
+	err = cmd.Wait()
+	if err != nil {
+		return fmt.Errorf("Fetching pull request: %v", err)
+	}
+	return nil
+}
+
+// Merge merges changes on the "_pr" branch
+func Merge() error {
```
jer: How are there no arguments here?

jer: Oh, it merges _pr? Yeah that's too magical. Merge should just merge what you tell it to merge, rather than assuming that some stuff happened before it was run.

git/git.go
```diff
@@ -22,3 +23,37 @@ func Clone(repo, destination string) error {
 	}
 	return nil
 }
+
+// Fetch fetches a pull request into a branch named "_pr"
+func Fetch(prNumber string) error {
```
jer: something called Fetch shouldn't accept a pr number. Something called FetchPR could, but not something called Fetch

launch.go
jer: definitely no need to mock out strings.Split. You can always control what gets sent to it

launch.go
```diff
@@ -88,6 +93,23 @@ func createWorkspace(srcPaths ...string) (Workspace, error) {
 	return w, nil
 }
 
+// checkPR checks to see if the job name is a pull request and returns its number
```
jer: Is the job name the best way to know? Can't we get that out of the SCM URL or something?

launch.go
jer: Sort of questionable. Ideally you would be able to push data into the function that calls this one and just let it do its thing.

launch.go
```diff
@@ -114,6 +136,21 @@ func launch(api screwdriver.API, buildID string) error {
 	if err != nil {
 		return err
 	}
+	prNumber, err := checkIfPr(j.Name)
```
jer: "check" to me doesn't imply returning the PR number. We should revisit this interface.

launch.go
```diff
@@ -88,6 +93,23 @@ func createWorkspace(srcPaths ...string) (Workspace, error) {
...
 }
 
+// checkPR checks to see if the job name is a pull request and returns its number
+func checkIfPR(jobName string) (string, error) {
+	prNumber := ""
+	reg := "^PR-[0-9]+$"
+
+	matched, err := matchString(reg, jobName)
+	if err != nil {
+		return prNumber, nil
```
jer: prNumber, nil?

launch.go
jer: Also prNumber seems like it should be a number. "branch" is more like a string.

launch_test.go
jer: I think it's fine to combine them when there is so little to them. I'd even make it more pointed like "TestMergePR"

launch.go
Filbird: Understood! Also choosing to not mock out regexp.MatchString then.

launch.go
Filbird: Hmm seems so, don't think SCM URL has an attribute indicating whether one of its jobs is a PR.

jer:
👍

nkatzman:
Seems fine 👍

jer:
👍  GREAT TYPING!

Issue #24, opened 2016-08-04
Clone over HTTPS


jer:
Currently, the launcher only works over SSH. We should switch to HTTPS as the default. Eventually we can add optional cloning over SSH with deploy keys.

PR #28, opened 2016-08-05
Use https as default url for git clone operation

Filbird:
- Adds a helper function that will return the url in https format.
- Adds functions to set the git user name and email
- Addresses #24

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -35,3 +35,21 @@ func Clone(scmURL, destination string) error {
 	}
 	return nil
 }
+
+//SetConfig sets up git configuration
+func SetConfig(setting, name string) error {
+	cmd := execCommand("git", "config", setting, name)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	err := cmd.Start()
+	if err != nil {
+		return fmt.Errorf("starting git config command: %v", err)
+	}
+
+	err = cmd.Wait()
+	if err != nil {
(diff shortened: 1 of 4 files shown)


git/git.go
```diff
@@ -32,3 +32,39 @@ func Clone(scmURL, destination string) error {
 	}
 	return nil
 }
+
+//SetUserName sets up git configuration user.name to the specified name
+func SetUserName(name string) error {
```
stjohnjohnson: Can we combine these methods?

stjohnjohnson: e.g. `SetConfig(field string, name string)`

launch.go
```diff
@@ -24,11 +25,18 @@ type scmPath struct {
...
 var gitClone = git.Clone
+var gitSetUserName = git.SetUserName
+var gitSetUserEmail = git.SetUserEmail
 
 func (s scmPath) String() string {
 	return fmt.Sprintf("%s:%s/%s#%s", s.Host, s.Org, s.Repo, s.Branch)
 }
 
+func (s scmPath) httpsString() string {
+	host := strings.Split(s.Host, "@")[1]
```
jer: If the url is github.com/screwdriver-cd/launcher, this will panic. At a minimum you should add @ to the regex below so that an error will get thrown if @ doesn't exist.

jer: Best would probably be to extract <user>@ as an optional field from the url.

jer:
👍

jer:
👍

PR #29, opened 2016-08-05
Some more enhancements

stjohnjohnson:
- Calling out an explicit dependency on /v3
- Checking out a desired branch
- Adding logging to startup and git clone
- Allow alternative workspace directory
- Support token loading from environment

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,18 @@
 
 ## Usage
 
+```bash
+$ go get github.com/screwdriver-cd/launcher
+$ launch --api-uri http://localhost:8080/v3 cba94a05f8aa063f4b8cfb62cbc355e0c5f02698
+```
+
+## Testing
+
+```bash
+$ go get github.com/screwdriver-cd/launcher
+$ go test -cover github.com/screwdriver-cd/launcher/...
+```
+
 ## License
 
 Code licensed under the BSD 3-Clause license. See LICENSE file for terms.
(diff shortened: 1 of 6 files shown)


Filbird:
👍

README.md
```diff
@@ -5,6 +5,19 @@
 
 ## Usage
 
+```bash
+$ go get github.com/screwdriver-cd/launcher
```
jer: Assuming $GOPATH/bin is in your PATH (it is supposed to be), this section can be shortened to:

```bash
go get github.com/screwdriver-cd/launcher
launch --api-uri http://localhost:8080/v3 cba94a05f8aa063f4b8cfb62cbc355e0c5f02698
```

launch.go
```diff
@@ -81,77 +81,101 @@ func createWorkspace(srcPaths ...string) (Workspace, error) {
...
 	if err != nil {
-		return err
+		return fmt.Errorf("creating workspace: %v", err)
 	}
 
-	err = gitClone(p.ScmURL, workspace.Src)
+	err = gitClone(p.ScmURL, w.Src)
 	if err != nil {
-		return err
+		return fmt.Errorf("cloning from git: %v", err)
```
jer: I don't feel like this Errorf or the one above add any information, since the error actually already says the same thing with more info. That's why we were returning nil. Perhaps log.Errorf(err) would be most appropriate here? That would exit immediately after writing the log message.

jer:
👍

PR #30, opened 2016-08-05
Added Screwdriver YAML parsing via the /validator endpoint

jer:
validator

PR #34, opened 2016-08-08
Incorporate api /validator call

nkatzman:
This PR ties the api function `PipelineDefFromYaml` defined in #30 to the launch function

Adds an executor library that simply returns nil and should be called with the job to be executed


jer:
👍

PR #40, opened 2016-08-11
All code paths should now lead to marking the build success/failure

jer:
Please pay special attention when reviewing to try to catch any possible exit that would not correctly set the build status

- The command should exit 0 no matter what, to simplify job execution in
  Kubernetes
- Build failures should give good log messages and set the status to
  FAILED
- Unexpected errors that are not non-zero exit codes should result in
  making the build failed
- A panic anywhere in the codebase should attempt to mark the build
  failed, write a stack trace, and exit(0)

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -8,8 +8,11 @@ import (
 	"log"
 	"os"
 	"path"
+	"path/filepath"
 	"regexp"
+	"runtime/debug"
 	"strings"
+	"time"
 
 	"github.com/screwdriver-cd/launcher/executor"
 	"github.com/screwdriver-cd/launcher/git"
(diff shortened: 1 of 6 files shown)


jer:
Currently doesn't do the right thing if it dies from a signal. I'll follow up with another PR on that one.

launch.go
```diff
@@ -20,20 +23,36 @@ import (
...
 var open = os.Open
 var executorRun = executor.Run
 var writeFile = ioutil.WriteFile
 
+var cleanExit = func() {
+	os.Exit(0)
+}
+
+// exit sets the build status and exits successfully
+func exit(status screwdriver.BuildStatus, api screwdriver.API) {
```
nkatzman: could this exit function ever have an unexpected error?

jer: I assume it would be a panic in that case, and the recoveries in launch and main should deal with them.

launch.go
```diff
@@ -213,32 +233,43 @@ func launch(api screwdriver.API, buildID string, rootDir string) error {
...
 }
 
+func recoverPanic(api screwdriver.API) {
+	if p := recover(); p != nil {
+		filename := fmt.Sprintf("launcher-stacktrace-%s", time.Now().Format(time.RFC3339))
+		tracefile := filepath.Join(os.TempDir(), filename)
+
+		log.Printf("ERROR: Internal Screwdriver error. Please file a bug about this: %v", p)
+		log.Printf("ERROR: Writing StackTrace to %s", tracefile)
+		err := ioutil.WriteFile(tracefile, debug.Stack(), 0600)
```
nkatzman: What happens if we can't write the file?

jer: Ok, I just pushed a change that will try one last time to output something useful before dying: finalRecover()

petey:
👍  Looks good from my extremely limited Go experience.

PR #45, opened 2016-08-15
Add git reset function

Filbird:
Calls "git reset --hard [sha]" after cloning to ensure HEAD points to correct commit.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -14,6 +14,7 @@ var setConfig = SetConfig
 var mergePR = MergePR
 var fetchPR = FetchPR
 var merge = Merge
+var reset = Reset
 
 // command executes the git command
 func command(arguments ...string) error {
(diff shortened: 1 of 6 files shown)


git/git.go
```diff
@@ -74,13 +75,23 @@ func MergePR(prNumber string, branch string) error {
...
 }
 
+//Reset resets the git HEAD to the specified commit
+func Reset(sha string) error {
+	return command("reset", "--hard", sha)
+}
+
 // Setup clones a repository, sets the local config, and merges a PR if necessary
-func Setup(scmURL, destination, pr string) error {
+func Setup(scmURL, destination, pr string, sha string) error {
```
jer: func Setup(scmURL, destination, pr, sha string) error {

jer:
a) I feel like the whole interface is wrong on this git stuff and needs to be dramatically refactored
b) 👍  (after the function signature is updated)

Filbird:
Hrmm. Would it be more idiomatic to define a Git interface similar to what we do for all the API calls?

PR #48, opened 2016-08-16
Fixing the tests to work with go 1.7

jer:
Apparently go 1.7 gets mad if you fail a test after the test is
complete. This means that our sloppy use of global function overrides
caught up with us.


nkatzman:
👍

PR #49, opened 2016-08-17
Refactor git package

Filbird:
Initial stab at refactoring the git package

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -9,111 +9,113 @@ import (
 )
 
 var execCommand = exec.Command
-var clone = Clone
-var setConfig = SetConfig
-var mergePR = MergePR
-var fetchPR = FetchPR
-var merge = Merge
-var reset = Reset
+
+const (
+	SCM_URL_INDEX = 0
+	BRANCH_INDEX  = 1
+)
+
+type repo struct {
+	ScmURL string
+	Path   string
+	Branch string
+}
+
+// Repo is a git repo
+type Repo interface {
+	Checkout() error
+	MergePR(prNumber, sha string) error
+	GetPath() string
+}
+
+// New returns a new repo object
+func New(scmURL, path string) (Repo, error) {
+	parts := strings.Split(scmURL, "#")
+	if len(parts) < 2 {
+		return nil, fmt.Errorf("expected #branchname in SCM URL: %v", scmURL)
+	}
+	repo := repo{
+		ScmURL: parts[SCM_URL_INDEX],
+		Path:   path,
+		Branch: parts[BRANCH_INDEX],
+	}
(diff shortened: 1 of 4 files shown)


launch.go
```diff
@@ -182,11 +182,23 @@ func launch(api screwdriver.API, buildID string, rootDir string) error {
...
 		return err
 	}
 
+	err = repo.Checkout(b.SHA)
+	if err != nil {
+		return err
+	}
+
+	if pr != "" {
+		err = repo.MergePR(pr, "pr")
```
jer: `"pr"` here seems unnecessary. Can we just move it into MergePR and make it a single parameter?

Filbird: Sounds good! I'm thinking we still want to ensure merge takes a branch argument, so something like the following:  
```
func (r repo) MergePR(prNumber string) error {
    branch := "pr"
	if err := fetchPR(prNumber, branch, r.Path); err != nil {
		return fmt.Errorf("fetching pr: %v", err)
	}

	if err := merge(branch, r.Path); err != nil {
		return fmt.Errorf("merging pr: %v", err)
	}
	return nil
}

```

jer: Yep, perfect!

git/git_test.go
```diff
@@ -97,417 +58,125 @@ func TestHelperProcess(*testing.T) {
...
-		want := []string{
-			"merge", "--no-edit", testBranch,
-		}
-		if len(args) != len(want) {
-			t.Errorf("Incorrect args sent to git: %q, want %q", args, want)
-		}
-		for i, arg := range args {
-			if arg != want[i] {
-				t.Errorf("args[%d] = %q, want %q", i, arg, want[i])
+			switch args[0] {
```
jer: Maybe add a default case of t.Errorf here?

Filbird: Yep, makes sense to do so.

git/git.go
```diff
@@ -9,111 +9,113 @@ import (
...
-// FetchPR fetches a pull request into a specified branch
-func FetchPR(prNumber string, branch string) error {
-	return command("fetch", "origin", "pull/"+prNumber+"/head:"+branch)
-}
-
-// Merge merges changes on the specified branch
-func Merge(branch string) error {
-	return command("merge", "--no-edit", branch)
+	repo := repo{
+		ScmURL: parts[0],
```
nkatzman: should we add const values for the 0 and 1 magic numbers?

Something like "SCM_URL_INDEX" and "BRANCH_INDEX"

Filbird: good catch! Will make the change

jer:
This is awesome 👍

PR #51, opened 2016-08-19
Add retry wrapper for http calls

Filbird:
- All HTTP calls will attempt to retry upon receiving 5XXs
- Default retry attempt is currently set to 5
- Exponential back off for up to ~1 minute (2s, 4s, 8s, 16s, 32s)

- Also refactors Screwdriver tests into tables
- Change const values into camelCase

### DIFF
--- a/screwdriver/screwdriver.go
+++ b/screwdriver/screwdriver.go
@@ -3,12 +3,18 @@ package screwdriver
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
+	"math"
 	"net/http"
 	"net/url"
+	"time"
 )
+
+var sleep = time.Sleep
 
 // BuildStatus is the status of a Screwdriver build
 type BuildStatus string
@@ -20,6 +26,8 @@ const (
 	Failure             = "FAILURE"
 	Aborted             = "ABORTED"
 )
+
+const maxAttempts = 5
 
 func (b BuildStatus) String() string {
 	return string(b)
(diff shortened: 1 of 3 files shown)


screwdriver/screwdriver.go
```diff
@@ -3,13 +3,20 @@ package screwdriver
...
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"net/url"
+	"time"
+
+	"github.com/jpillora/backoff"
```
jer: I don't want to use a library for this. It's too easy to write a retry wrapper and not worth pulling a dependency in.

screwdriver/screwdriver.go
```diff
@@ -137,34 +144,102 @@ func handleResponse(res *http.Response) ([]byte, error) {
...
-	res, err := a.client.Do(req)
+	res := &http.Response{}
+	resError := errors.New("")
+	err = retry(5, func() error {
+		res, resError = a.client.Do(req)
+		if resError != nil {
+			return nil
+		}
+		if res.StatusCode/100 == 5 {
+			return fmt.Errorf("response contains status code %v, retrying GET request", res.StatusCode)
```
jer: This isn't entirely accurate, is it? After 5 errors, iIt would result in something like this:

"timeout on request: response contains status code 500, retying GET request"

Would be nicer to have 2 things:

1) After failure, log.Printf("WARNING: received response 500 from GET http://api.screwdriver.cd/builds/123 retrying (1/5)")
2) When all requests are exhausted, return error:

"GET retries exhausted: 500 returned from GET http://api.screwdriver.cd/builds/123"

screwdriver/screwdriver.go
```diff
@@ -21,6 +27,8 @@ const (
 	Aborted             = "ABORTED"
 )
 
+const MAX_ATTEMPTS = 5
```
jer: Didn't catch this the first pass, but Const in Go is generally CamelCase

jer:
👍  other than the capitalization thing

screwdriver/screwdriver.go
Filbird: On the path to being a good Go citizen. Will change the other consts as well 😄

jer:
👍

PR #52, opened 2016-08-19
Change dir of each command, Merge environment variables instead of overrwite

Filbird:
Rather than changing the directory in launch, we do so in executor.Run by passing in the path to the function.

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -3,6 +3,7 @@ package executor
 import (
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
 	"syscall"
 
@@ -21,18 +22,20 @@ func (e ErrStatus) Error() string {
 }
 
 // Run executes a slice of CommandDefs
-func Run(output io.Writer, job screwdriver.JobDef) error {
+func Run(path string, output io.Writer, job screwdriver.JobDef) error {
 	cmds := job.Commands
-	env := job.Environment
+	env := os.Environ()
+	for k, v := range job.Environment {
+		env = append(env, fmt.Sprintf("%s=%s", k, v))
+	}
 	for _, cmd := range cmds {
 		shargs := []string{"-e", "-c"}
 		shargs = append(shargs, cmd.Cmd)
 		c := execCommand("sh", shargs...)
 		c.Stdout = output
 		c.Stderr = output
-		for k, v := range env {
-			c.Env = append(c.Env, fmt.Sprintf("%s=%s", k, v))
-		}
+		c.Dir = path
+		c.Env = append(env, c.Env...)
 
 		if err := c.Start(); err != nil {
 			return fmt.Errorf("launching command %q: %v", cmd.Cmd, err)
--- a/launch_test.go
+++ b/launch_test.go
@@ -24,6 +24,7 @@ type FakePipelineDef screwdriver.PipelineDef
 type MockRepo struct {
 	checkout func() error
 	mergePR  func(string, string) error
+	getPath  func() string
 }
 
 func (r MockRepo) Checkout() error {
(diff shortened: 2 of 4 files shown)


stjohnjohnson:
So every command does a chdir before?

Filbird:
Believe it will only occur once for every job =)

stjohnjohnson:
LGTM 👍

executor/executor.go
```diff
@@ -21,7 +23,10 @@ func (e ErrStatus) Error() string {
 }
 
 // Run executes a slice of CommandDefs
-func Run(output io.Writer, job screwdriver.JobDef) error {
+func Run(path string, output io.Writer, job screwdriver.JobDef) error {
+	if err := chdir(path); err != nil {
```
jer: Why are we doing this instead of just passing it as the working directory for execute?(os/exec.cmd.Dir)

Like this: https://github.com/screwdriver-cd/launcher/blob/master/git/git.go#L111

Filbird: Good point. Seems like setting the dir for each command would be more
reliable, especially since commands could change the working directory.
Will change :)

On Aug 20, 2016 8:21 PM, "Jeremiah Wuenschel" <notifications@github.com>
wrote:
>
> In executor/executor.go:
>
> > @@ -21,7 +23,10 @@ func (e ErrStatus) Error() string { > } > > // Run
executes a slice of CommandDefs > -func Run(output io.Writer, job
screwdriver.JobDef) error { > +func Run(path string, output io.Writer, job
screwdriver.JobDef) error { > + if err := chdir(path); err != nil {
>
> Why are we doing this instead of just passing it as the working directory
for execute?(os/exec.cmd.Dir)
>
> Like this:
https://github.com/screwdriver-cd/launcher/blob/master/git/git.go#L111

stjohnjohnson: So does that prevent people from changing the directory in step 1, then using that new dir in step 2?

jer: Yes, they would have to chdir each step. What do we want to happen here? We should decide if we want it to act like a single shell for all steps or independent and consistent environments.

What I don't like is the introduction of a side-effect by the executor. Calling a helper function in a package shouldn't result in the caller having a different working directory.

It also seems awkward to have some shell side-effects but not others. If we let the directory change in a step and carry over, we should probably try to propagate the environment as well - like current Screwdriver.

jer: For reference: environment and cwd propagate between steps in Wercker.

stjohnjohnson: I would love both to carry over.  Similar to Wercker and Travis

jer: Actually, cd in the subshell does not carry over at all right now, regardless of where we chdir or pass in a path. Each call to exec sets it to CWD for the launch process by default.

Filbird: To address the concern of changing directories from a helper function call,
I believe we can just reset the working directory once all commands have
been executed.

As for propagating working directories, perhaps we can track the directory
at the end of each command execution and set  the dirs of subsequent calls

Filbird:
We are now setting the dir for each command to execute from the specified path. As of now, commands that may change the working directory will not propagate that change to subsequent commands.

Also addressing #53 in that each command retains the default environment variable of the container whilst appending any environment variables defined in the Job.

executor/executor.go
```diff
@@ -21,18 +22,20 @@ func (e ErrStatus) Error() string {
...
 		shargs := []string{"-e", "-c"}
 		shargs = append(shargs, cmd.Cmd)
 		c := execCommand("sh", shargs...)
 		c.Stdout = output
 		c.Stderr = output
-		for k, v := range env {
-			c.Env = append(c.Env, fmt.Sprintf("%s=%s", k, v))
-		}
+		c.Dir = path
+		c.Env = append(env, c.Env...)
```
jer: This should technically work because later env variables will override earlier ones, but it would be cleaner to de-dupe things in a future update

launch_test.go
```diff
@@ -24,6 +24,7 @@ type FakePipelineDef screwdriver.PipelineDef
 type MockRepo struct {
 	checkout func() error
 	mergePR  func(string, string) error
+	getPath  func() string
```
jer: can we call this "path()" instead?

jer:
Seems fine. We should refactor the repo path stuff a little bit i think

jer:
👍

launch_test.go
Filbird: For just the MockRepo or for all Repos?
In the latter case we may run into an issue where we have "Path" defined as both a field and a method.

PR #56, opened 2016-08-23
Copying the payload to a string for POST retry

jer:
- This was causing retries to contain an empty payload


Filbird:
👍

nkatzman:
👍  seems reasonable

PR #58, opened 2016-08-24
Add default environment variables in launch

Filbird:
Adds the environment variables specified in #57 with their default values.

Issue #57, opened 2016-08-23
Default Environment Variables


stjohnjohnson:
All CI systems have environment variables built in to help the user.  We should do the same:

 - SCREWDRIVER: true
 - CI: true
 - CONTINUOUS_INTEGRATION: true
 - SD_JOB_NAME: job name
 - SD_PULL_REQUEST: pull request number
 - SD_SOURCE_DIR: directory that the code was checked out in
 - SD_ARTIFACTS_DIR: directory for artifacts

Existing examples:
 - http://devcenter.wercker.com/docs/environment-variables/available-env-vars.html
 - https://docs.travis-ci.com/user/environment-variables

stjohnjohnson:
works for me

PR #59, opened 2016-08-24
Add default environment variables as argument to executor run

Filbird:
Alternative method to add default environment variables (#57) , please compare to PR #58 

My thoughts are that this implementation moves the responsibility of setting the default environment variables from launch to the executor.


stjohnjohnson:
LGTM 👍

stjohnjohnson:
```
./launch.go:250: too many arguments in call to executorRun
```

stjohnjohnson:
test

PR #61, opened 2016-08-26
Fixed a bug with partial line reads in the emitter

jer:
- Previously reads were happening in a buffer, which could lead to
  partial lines instead of waiting for newline
- Now we are using a Pipe to allow us to pull in STDOUT/STDERR from the command subprocesses.
  This has the effect of creating a blocking pipe all the way through to the unix socket that we are 
  writing to. We may need to buffer things in the future if we start to hit delays writing to the
  logger service over the unix socket

### DIFF
--- a/executor/executor_test.go
+++ b/executor/executor_test.go
@@ -37,6 +37,10 @@ type MockEmitter struct {
 	found    []byte
 }
 
+func (e *MockEmitter) Error() error {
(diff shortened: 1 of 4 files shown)


Filbird:
👍 📖 ✏️

stjohnjohnson:
giphy 👍

PR #62, opened 2016-08-26
Adding the command that is being executed to the Emitter output

jer:
- Also changed to mS instead of μS in the log times


d2lam:
👍  for fancy U

stjohnjohnson:
👎  not enough 
giphy 2

PR #65, opened 2016-08-30
Bumping the tag in the deploy step

jer:
We need this because the repo isn't reused between build and deploy


tkyi:
👍 seems like a good idea

PR #69, opened 2016-09-01
Adding Dockerfile and fixing wercker.yml

stjohnjohnson:
Moving from Job-Tools


tkyi:
👍

Filbird:
So in the job-tools init container specified in https://github.com/screwdriver-cd/executor-k8s/blob/master/config/job.yaml.tim, will the job-tools image be pulling from `"screwdrivercd/launcher:{{launch_version}}"` instead of `"screwdrivercd/job-tools:{{tools_version}}"`?

PR #71, opened 2016-09-06
Associating emitter data with sd-setup by default

jer:
This will allow us to emit information about the Screwdriver workspace
setup, and have it be associated with the sd-setup test


stjohnjohnson:
what

LGTM 👍

PR #74, opened 2016-09-09
Improved environment population

jer:
- If the container has an environment like this:
    FOO="bar=baz"
  we now set the correct value
- Added a test for environment generation


FenrirUnbound:
👍 

Looks good. Unsure why the Screwdriver build failed

tumblr_oczgi8ojeh1vf8r3co2_r1_540

stjohnjohnson:
👍 
equal

PR #83, opened 2016-10-04
Improving http retries

jer:
- Added a client timeout of 10 seconds
- Retries now happen for things like eof and timeout that have no http
  response code
- Cleaned up some confusing log statements

### DIFF
--- a/screwdriver/screwdriver.go
+++ b/screwdriver/screwdriver.go
@@ -3,7 +3,6 @@ package screwdriver
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


screwdriver/screwdriver.go
```diff
@@ -195,30 +194,28 @@ func (a api) get(url *url.URL) ([]byte, error) {
...
+		attemptNumber++
+		res, err = a.client.Do(req)
+		if err != nil {
+			log.Printf("WARNING: received error from GET(%s): %v "+
+				"(attempt %d of %d)", url.String(), err, attemptNumber, maxAttempts)
+			return err
 		}
+
 		if res.StatusCode/100 == 5 {
 			log.Printf("WARNING: received response %v from GET %v "+
```
d2lam: Should this be `GET %s` since it's `url.String()`?

PR #85, opened 2016-10-11
Adding Fifo and Log Service

stjohnjohnson:
Also switching to web instead of API calls (as it breaks in Docker Hub sometimes due to rate limiting)


jer:
👍  lookin good

PR #87, opened 2016-10-13
Use screwdriver and remove wercker

minz1027:


jer:
👍  looks good.

All the duplication kind of sucks, but we need to be able to save artifacts in jobs to do anything about that.

PR #90, opened 2016-10-20
Update parseScmUrl() to take in scmUri and scmRepo

minz1027:
- Change `parseScmUrl()` to `parseScmUri()`
- Now the method takes in `scmUri`, `scmRepo.name`  and `scmRepo.branch`

screen shot 2016-10-19 at 8 08 20 pm

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -54,27 +54,24 @@ type scmPath struct {
 	Branch string
 }
 
-func (s scmPath) String() string {
-	return fmt.Sprintf("git@%s:%s/%s#%s", s.Host, s.Org, s.Repo, s.Branch)
-}
-
 func (s scmPath) httpsString() string {
 	return fmt.Sprintf("https://%s/%s/%s#%s", s.Host, s.Org, s.Repo, s.Branch)
 }
 
-// e.g. "git@github.com:screwdriver-cd/launch.git#master"
-func parseScmURL(url string) (scmPath, error) {
-	r := regexp.MustCompile("^git@(.*):(.*)/(.*?)(?:\\.git)?#(.*)$")
-	matched := r.FindStringSubmatch(url)
-	if matched == nil || len(matched) != 5 {
-		return scmPath{}, fmt.Errorf("unable to parse SCM URL %v, match: %q", url, matched)
+// e.g. scmUri: "github:123456:master", scmName: "screwdriver-cd/launcher"
+func parseScmUri(scmUri, scmName string) (scmPath, error) {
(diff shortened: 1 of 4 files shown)


launch.go
```diff
@@ -180,7 +176,7 @@ func launch(api screwdriver.API, buildID, rootDir, emitterPath string) error {
 		return fmt.Errorf("fetching Pipeline ID %q: %v", j.PipelineID, err)
 	}
 
-	scm, err := parseScmURL(p.ScmURL)
+	scm, err := parseScmUri(p.ScmUri, p.ScmRepo.Name, p.ScmRepo.Branch)
```
d2lam: Don't think we need to pass in branch, it could be `uri[2]`

minz1027: Fixed

launch.go
```diff
@@ -54,27 +54,23 @@ type scmPath struct {
...
 	return fmt.Sprintf("https://%s/%s/%s#%s", s.Host, s.Org, s.Repo, s.Branch)
 }
 
-// e.g. "git@github.com:screwdriver-cd/launch.git#master"
-func parseScmURL(url string) (scmPath, error) {
-	r := regexp.MustCompile("^git@(.*):(.*)/(.*?)(?:\\.git)?#(.*)$")
-	matched := r.FindStringSubmatch(url)
-	if matched == nil || len(matched) != 5 {
-		return scmPath{}, fmt.Errorf("unable to parse SCM URL %v, match: %q", url, matched)
+func parseScmUri(scmUri, scmName, scmBranch string) (scmPath, error) {
```
tkyi: We were worried the `url` would be different for github vs some other scm and thought using the scmUri, name, and branch would be a more reliable choice

screwdriver/screwdriver.go
```diff
@@ -96,7 +96,13 @@ type Validator struct {
 // Pipeline is a Screwdriver Pipeline definition.
 type Pipeline struct {
 	ID     string `json:"id"`
-	ScmURL string `json:"scmUrl"`
+	ScmRepo ScmRepo `json:"scmRepo"`
+	ScmUri string `json:"scmUri"`
+}
+
+// ScmRepo is an object inside a Pipeline
```
jer: Not really. It's more of a struct containing a Name than an object that is somewhere. But what is an SCM Repo?

minz1027: Fixed!
The new API will have a `scmRepo` object inside the pipeline.
https://beta.api.screwdriver.cd/v4/pipelines

PR #92, opened 2016-11-08
Get commands and environment from build instead of job

tkyi:
We can get the commands and environment of a build from a build instead of a job for more up to date information


d2lam:
Let's build a test image and run it locally before merging.

minz1027:
Tested locally with in-a-box , works! Image is built based on this PR with the tag `testGetConfig`
https://hub.docker.com/r/screwdrivercd/launcher/tags/

d2lam:
Yup, I see it working!

PR #95, opened 2016-11-18
Refactor build environment creation

d2lam:
- Setting build environment variables can be done inside `createEnvironment` since it just reads from `build`. 
- Remove some unused code left from the previous refactoring

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -216,7 +216,7 @@ func launch(api screwdriver.API, buildID, rootDir, emitterPath string) error {
 		return fmt.Errorf("Fetching secrets for build %s", b.ID)
 	}
 
-	env := createEnvironment(defaultEnv, secrets)
+	env := createEnvironment(defaultEnv, secrets, b)
 
 	if err := api.UpdateStepStop(buildID, "sd-setup", 0); err != nil {
 		return fmt.Errorf("updating sd-setup stop: %v", err)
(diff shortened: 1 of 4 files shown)


launch_test.go
```diff
@@ -40,27 +40,6 @@ type MockRepo struct {
...
-}
-
-func (r MockRepo) Path() string {
-	if r.path != nil {
-		return r.path()
-	}
-	return ""
-}
-
 func mockAPI(t *testing.T, testBuildID, testJobID, testPipelineID string, testStatus screwdriver.BuildStatus) MockAPI {
```
minz1027: Should also remove https://github.com/screwdriver-cd/launcher/blob/79990efb37df711f02fda7a3424de44c802258f4/launch_test.go#L37

d2lam: Good catch! removed.

launch.go
```diff
@@ -268,6 +268,11 @@ func createEnvironment(base map[string]string, secrets screwdriver.Secrets) []st
 	for k, v := range combined {
 		envStrings = append(envStrings, strings.Join([]string{k, v}, "="))
 	}
+
+	for k, v := range build.Environment {
```
petey: We should probably have some discussion about what env variables we do, and what we don't want the user to have influence over. This would allow the user to override all the default env settings in their sd.yaml if they wanted to.

d2lam: Hmm yeah, we should. But this PR doesn't change any previous behaviors, it just refactors.

stjohnjohnson: Right, shouldn't we be adding our SD environment variables in the config-parser?  That way we can control which order they apply in and what they are.

d2lam: @stjohnjohnson We can discuss about that more. But this PR doesn't change the behaviors. It has always been adding ENV this way inside the launcher. This is to refactor/clean up code.

stjohnjohnson: Well okay then.

PR #96, opened 2016-11-23
[major] fix: change volume for launcher

petey:
Prevent overriding volumes in legacy containers.


d2lam:
Where & how do we bump version?

PR #97, opened 2016-11-24
feat: Switching to new naming convention `sd-setup-*` for startup commands [minor]

stjohnjohnson:
In the API we're naming all setup steps with `sd-setup-`

PRs needed:
 - https://github.com/screwdriver-cd/build-bookend/pull/4
 - https://github.com/screwdriver-cd/models/pull/143


d2lam:
Don't you need to change models too?

stjohnjohnson:
Actually, I'd love to remove this completely if it's not helpful.  @jer do you remember why we have https://github.com/screwdriver-cd/launcher/blob/cecfa6009b4c289dd40581032e30f384998543ba/screwdriver/emitter.go#L73-L75

tkyi:
Is this naming convention documented somewhere?

d2lam:
@stjohnjohnson  It's to log the thing that's happening here: https://github.com/screwdriver-cd/launcher/blob/ad07ecf62879d49884e3bf99b64ee9530176258e/launch.go#L140-L223

Creating workspace and setting up environments don't belong to any steps. If we don't have it, then how do we log it?

petey:
config-parser will also need updates to prevent a user from making a `setup-*` or `teardown-*` step. The parser only prevents `sd-*`.

https://github.com/screwdriver-cd/config-parser/blob/98f8b11dd67943d425d880afb80e94221a9d95b5/lib/phase/functional.js#L64-L69

PR #101, opened 2017-01-19
Launcher passes env variables between steps

jerryzhang222:
Writing each command in the YAML to a temporary file in launcher. Instead of running the commands from YAML, source the temporary file to run the command.

Related to: screwdriver-cd/screwdriver#393

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -1,23 +1,27 @@
 package executor
 
 import (
+	"bufio"
 	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
 	"os/exec"
-	"syscall"
+	"regexp"
+	"strconv"
+	"strings"
 
+	"github.com/kr/pty"
+	"github.com/myesui/uuid"
 	"github.com/screwdriver-cd/launcher/screwdriver"
 )
 
 const (
-	// ExitLaunch is the exit code when a step fails to launch
-	ExitLaunch = 255
 	// ExitUnknown is the exit code when a step doesn't return an exit code (for some weird reason)
 	ExitUnknown = 254
 	// ExitOk is the exit code when a step runs successfully
 	ExitOk = 0
 )
-
-var execCommand = exec.Command
 
 // ErrStatus is an error that holds an exit status code
 type ErrStatus struct {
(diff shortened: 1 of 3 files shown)


executor/executor.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"os/exec"
 	"syscall"
+  "io/ioutil"
```
stjohnjohnson: Please make sure you have `gofmt` installed as it will auto-indent for you on save.

executor/executor.go
```diff
@@ -28,45 +33,104 @@ func (e ErrStatus) Error() string {
...
-	c.Stdout = emitter
-	c.Stderr = emitter
+	f.Write([]byte(shargs))
 
+	return copyLinesUntil(fReader, emitter, guid)
+}
+
+// Run executes a slice of CommandDefs
+func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriver.Build, api screwdriver.API, buildID int) error {
+	// Set up a single psydo-terminal
```
d2lam: typo

d2lam:
@jer  Can you take a look at this PR?

executor/executor.go
```diff
@@ -1,24 +1,28 @@
...
+	"io"
+	"io/ioutil"
+	"os"
 	"os/exec"
-	"syscall"
+	"regexp"
+	"strconv"
+	"strings"
 
+	"github.com/kr/pty"
```
jer: We should use glide if we are going to depend on an external package. That will prevent us from immediately breaking when master breaks upstream.

executor/executor.go
```diff
@@ -28,53 +32,114 @@ func (e ErrStatus) Error() string {
...
-	fmt.Fprintf(emitter, "$ %s\n", cmd.Cmd)
-	c.Stdout = emitter
-	c.Stderr = emitter
+// Copy lines until match string
+func copyLinesUntil(r io.Reader, w io.Writer, match string) (int, error) {
+	scanner := bufio.NewScanner(r)
+	for scanner.Scan() {
+		t := scanner.Text()
+		// Match the guid and exitCode
+		re := regexp.MustCompile("(" + match + ") " + "([0-9]+)")
```
jer: I think it's unusual to concatenate strings this way. Would this approach work?

```go
regex := fmt.Sprintf("(%s) ([0-9]+)", match)
re := regexp.MustCompile(regex)
```

executor/executor.go
d2lam: Hey @jer, couldn't get it to work. Having same issues as these people here:
https://github.com/Masterminds/glide/issues/298
https://github.com/Masterminds/glide/issues/577

Those issues are still opened. Also, looks like kr/pty is not being developed. Only 1 release so far. Maybe we can make this work in the next PR?

jer:
Okaaaaaay 👍

d2lam:
Waiting to integrate one more time to make sure everything works

PR #105, opened 2017-02-16
Give all users RW permission to /opt/sd

r3rastogi:
/opt/sd is currently owned by root and when we try to write to it as any
other user (for example, to create the emitter directory), we run into
permission denied errors. We want Screwdriver to be able to be run as a
non-root user, so we need to give all users permissions to this
directory.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@ RUN set -x \
    && mv tini-static tini \
    && chmod +x tini \
    # Create FIFO
-   && mkfifo emitter \
+   && mkfifo -m 666 emitter \
    # Cleanup packages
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -34,6 +34,7 @@ RUN set -x \
    && rm tini-static.asc \
    && mv tini-static tini \
    && chmod +x tini \
+   && chmod 777 /opt/sd \
    # Create FIFO
    && mkfifo emitter \
```
stjohnjohnson: Can we just do `-m 666` to give everyone write permission?

r3rastogi: changed!

jer:
👍

PR #108, opened 2017-02-23
feat: using meta-cli in launcher

tk3fftk:
To use meta-cli, `meta` of previous build is load from API and stored to temporary file before a build starting.
Also current `meta` is added to API payload with `UpdateBuildStatus`.

Related: https://github.com/screwdriver-cd/screwdriver/issues/293

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,11 @@ RUN set -x \
       | egrep -o '/screwdriver-cd/log-service/releases/download/v[0-9.]*/logservice' \
       | wget --base=http://github.com/ -i - -O logservice \
    && chmod +x logservice \
+   # Download Meta CLI
+   && wget -q -O - https://github.com/screwdriver-cd/meta-cli/releases/latest \
+      | egrep -o '/screwdriver-cd/meta-cli/releases/download/v[0-9.]*/meta-cli' \
+      | wget --base=http://github.com/ -i - -O meta-cli\
+   && chmod +x meta-cli\
    # Download Tini Static
    && wget -q -O - https://github.com/krallin/tini/releases/latest \
       | egrep -o '/krallin/tini/releases/download/v[0-9.]*/tini-static' \
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,9 +23,9 @@ RUN set -x \
    && chmod +x logservice \
    # Download Meta CLI
    && wget -q -O - https://github.com/screwdriver-cd/meta-cli/releases/latest \
-      | egrep -o '/screwdriver-cd/meta-cli/releases/download/v[0-9.]*/meta-cli' \
-      | wget --base=http://github.com/ -i - -O meta-cli\
-   && chmod +x meta-cli\
+      | egrep -o '/screwdriver-cd/meta-cli/releases/download/v[0-9.]*/meta' \
+      | wget --base=http://github.com/ -i - -O meta\
+   && chmod +x meta\
    # Download Tini Static
    && wget -q -O - https://github.com/krallin/tini/releases/latest \
       | egrep -o '/krallin/tini/releases/download/v[0-9.]*/tini-static' \
--- a/launch.go
+++ b/launch.go
@@ -27,17 +27,33 @@ var stat = os.Stat
 var open = os.Open
 var executorRun = executor.Run
 var writeFile = ioutil.WriteFile
+var readFile = ioutil.ReadFile
 var newEmitter = screwdriver.NewEmitter
+var unmarshal = json.Unmarshal
 
 var cleanExit = func() {
 	os.Exit(0)
 }
 
 // exit sets the build status and exits successfully
-func exit(status screwdriver.BuildStatus, buildID int, api screwdriver.API) {
+func exit(status screwdriver.BuildStatus, buildID int, api screwdriver.API, metaSpace string) {
 	if api != nil {
+		var metaInterface map[string]interface{}
+
+		log.Printf("Loading meta from %q/meta.json", metaSpace)
+		metaJson, err := readFile(metaSpace + "/meta.json")
+		if err != nil {
+			log.Printf("Failed to load %q/meta.json: %v", metaSpace, err)
+			metaJson = nil
+		} else {
+			err = unmarshal(metaJson, &metaInterface)
+			if err != nil {
+				log.Printf("Failed to Load %q/meta.json: %v", metaSpace, err)
+			}
+			metaJson = nil
+		}
 		log.Printf("Setting build status to %s", status)
-		if err := api.UpdateBuildStatus(status, buildID); err != nil {
+		if err := api.UpdateBuildStatus(status, metaInterface, buildID); err != nil {
 			log.Printf("Failed updating the build status: %v", err)
 		}
 	}
@@ -111,6 +127,14 @@ func createWorkspace(rootDir string, srcPaths ...string) (Workspace, error) {
 	return w, nil
 }
 
+var createMetaSpace = func(metaSpace string) error {
+	err := mkdirAll(metaSpace, 0777)
+	if err != nil {
+		return fmt.Errorf("Cannot create meta-space path %q: %v", metaSpace, err)
+	}
+	return nil
+}
+
 func writeArtifact(aDir string, fName string, artifact interface{}) error {
 	data, err := json.MarshalIndent(artifact, "", strings.Repeat(" ", 4))
 	if err != nil {
@@ -137,7 +161,7 @@ func prNumber(jobName string) string {
 	return matched[1]
 }
 
-func launch(api screwdriver.API, buildID int, rootDir, emitterPath string) error {
+func launch(api screwdriver.API, buildID int, rootDir, emitterPath string, metaSpace string) error {
 	emitter, err := newEmitter(emitterPath)
 	if err != nil {
 		return err
@@ -149,7 +173,7 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath string) error
 	}
 
 	log.Print("Setting Build Status to RUNNING")
-	if err = api.UpdateBuildStatus(screwdriver.Running, buildID); err != nil {
+	if err = api.UpdateBuildStatus(screwdriver.Running, nil, buildID); err != nil {
 		return fmt.Errorf("updating build status to RUNNING: %v", err)
 	}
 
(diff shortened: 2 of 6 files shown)


launch.go
```diff
@@ -149,7 +172,7 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath string) error
 	}
 
 	log.Print("Setting Build Status to RUNNING")
-	if err = api.UpdateBuildStatus(screwdriver.Running, buildID); err != nil {
+	if err = api.UpdateBuildStatus(screwdriver.Running, nil, buildID); err != nil {
```
tk3fftk: `nil` makes JSON as `{"meta":null}` when `json.Unmarshal` is executed.

launch.go
```diff
@@ -27,17 +27,32 @@ var stat = os.Stat
...
 // exit sets the build status and exits successfully
-func exit(status screwdriver.BuildStatus, buildID int, api screwdriver.API) {
+func exit(status screwdriver.BuildStatus, buildID int, api screwdriver.API, metaSpace string) {
 	if api != nil {
+		var metaInterface map[string]interface{}
+
+		log.Printf("Loading meta for setting build meta from %q/meta.json", metaSpace)
+		metaJson, err := readFile(metaSpace + "/meta.json")
+		if err != nil {
+			log.Printf("Failed to Load %q/meta.json: %v", metaSpace, err)
```
d2lam: probably lower case "L" here

screwdriver/screwdriver.go
```diff
@@ -73,7 +73,8 @@ func New(url, token string) (API, error) {
 
 // BuildStatusPayload is a Screwdriver Build Status payload.
 type BuildStatusPayload struct {
-	Status string `json:"status"`
+	Status string                 `json:"status"`
+	Meta   map[string]interface{} `json:"meta"`
```
tk3fftk: `meta` is defined in data-scheme: 
https://github.com/screwdriver-cd/data-schema/blob/master/models/build.js#L151

launch.go
```diff
@@ -27,17 +27,32 @@ var stat = os.Stat
...
 	os.Exit(0)
 }
 
 // exit sets the build status and exits successfully
-func exit(status screwdriver.BuildStatus, buildID int, api screwdriver.API) {
+func exit(status screwdriver.BuildStatus, buildID int, api screwdriver.API, metaSpace string) {
 	if api != nil {
+		var metaInterface map[string]interface{}
+
+		log.Printf("Loading meta for setting build meta from %q/meta.json", metaSpace)
```
d2lam: This log is confusing to me. Maybe just `Loading meta from %q/meta.json`

launch.go
```diff
@@ -171,6 +194,24 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath string) error
 		return fmt.Errorf("fetching Pipeline ID %d: %v", j.PipelineID, err)
 	}
 
+	log.Printf("Fetching Parent Build %d for Meta", b.ParentBuildID)
+	pb, err := api.BuildFromID(b.ParentBuildID)
+	if err != nil {
+		return fmt.Errorf("fetching Parent Build ID %d: %v", ParentBuildID, err)
```
d2lam: Probably should make these comments lowercase and consistent for easier reading. Fetching parent build ID:

launch.go
```diff
@@ -171,6 +194,24 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath string) error
 		return fmt.Errorf("fetching Pipeline ID %d: %v", j.PipelineID, err)
 	}
 
+	log.Printf("Fetching Parent Build %d for Meta", b.ParentBuildID)
+	pb, err := api.BuildFromID(b.ParentBuildID)
```
d2lam: Where is this `pb` used?

launch.go
d2lam: Also shouldn't this be `b.ParentBuildID`?

launch.go
```diff
@@ -171,6 +194,24 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath string) error
...
+	}
+
+	log.Printf("Creating Meta Space in %d", metaSpace)
+	err = createMetaSpace(metaSpace)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("Fetching Parent Build Meta JSON %d", b.Meta)
+	err = ioutil.WriteFile(metaSpace+"/meta.json", b.Meta, 0666)
```
d2lam: Should this be `pb.Meta`?

launch.go
tk3fftk: fixed..

launch.go
tk3fftk: fixed..

launch.go
tk3fftk: You are right. Thank you. I fixed.

launch.go
tk3fftk: It seems that these are lowercase comments and uppercase comments.
Does `log` use uppercase and `error` use lowercase?

fixed to `pb.ParentBuildID`.

launch.go
tk3fftk: fixed.

tk3fftk:
Updated tests and CI became green!

launch.go
```diff
@@ -171,6 +195,24 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath string) error
...
+	}
+
+	log.Printf("Creating Meta Space in %v", metaSpace)
+	err = createMetaSpace(metaSpace)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("Fetching Parent Build Meta JSON %v", pb.Meta)
+	err = writeFile(metaSpace+"/meta.json", []byte(pb.Meta), 0666)
```
catto: spaces around `+`

launch.go
```diff
@@ -171,6 +195,24 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath string) error
 		return fmt.Errorf("fetching Pipeline ID %d: %v", j.PipelineID, err)
 	}
 
+	log.Printf("Fetching Parent Build %d for Meta", b.ParentBuildID)
```
minz1027: The first job `main` and `pr` jobs will not have the `parentBuild Id`. But we shouldn't treat them as errors. should probably just skip fetching parent meta in those cases.

launch.go
```diff
@@ -111,6 +127,14 @@ func createWorkspace(rootDir string, srcPaths ...string) (Workspace, error) {
 	return w, nil
 }
 
+var createMetaSpace = func(metaSpace string) error {
+	err := mkdirAll(metaSpace, 0777)
```
minz1027: Why it need to be 777? We only need to read or write right?

launch.go
tk3fftk: `gofmt` makes it.
https://play.golang.org/p/W2JmdQSNns

launch.go
tk3fftk: I see. I did not care about the first job. In the case of first job, `parentBuild Id` is nil or other value?

launch.go
catto: 🙆

launch.go
d2lam: I think it will not have parentBuildId. 
Example here: 1272 triggers 1273
https://api.screwdriver.cd/v4/builds/1272 does not have parentBuildId
https://api.screwdriver.cd/v4/builds/1273 has parentBuildId as 1272

launch.go
d2lam: I think it wasn't consistent from the start when we wrote the code.

d2lam:
I think this looks good now beside Min's comment on the `ParentBuildId` for the first job. 👍

launch.go
tk3fftk: Thank you for show examples! 
I think the value which is defined but not initialize in Go is set to `0`. `b.ParentBuildiD` is maybe 0 when the build does not have `prentBuildId`. The skip condition should be 0.

tk3fftk: Fixed.

d2lam: @tk3fftk  Can you point me to where it's setting to `0`? I can't seem to find it. Thanks!

tk3fftk: @d2lam The Golang provides [zero-value](https://golang.org/ref/spec#The_zero_value). 
> When storage is allocated for a variable, either through a declaration or a call of new, or when a new value is created, either through a composite literal or a call of make, and no explicit initialization is provided, the variable or value is given a default value

Please run https://play.golang.org/p/t545Ir6EHf to see `ParentBuildID` is setting to `0`.

d2lam: Oh I see. I got it. Thanks @tk3fftk!

tk3fftk:
Thank you for the reviews!!

d2lam:
Adding `blocked` status. We should merge the `meta-cli` first.

Dockerfile
```diff
@@ -21,6 +21,11 @@ RUN set -x \
       | egrep -o '/screwdriver-cd/log-service/releases/download/v[0-9.]*/logservice' \
       | wget --base=http://github.com/ -i - -O logservice \
    && chmod +x logservice \
+   # Download Meta CLI
+   && wget -q -O - https://github.com/screwdriver-cd/meta-cli/releases/latest \
+      | egrep -o '/screwdriver-cd/meta-cli/releases/download/v[0-9.]*/meta-cli' \
+      | wget --base=http://github.com/ -i - -O meta-cli\
```
minz1027: @tk3fftk Hi Hiroki, when u release the `meta-cli`, you build the binary with the name `meta`
 https://github.com/screwdriver-cd/meta-cli/releases

tk3fftk: Oh I missed it. Fixed.

tk3fftk:
Should I `squash` before your merging?

launch.go
minz1027: @tk3fftk this will fail the validation when hitting the build update api since we expect meta to be an object. 
https://github.com/screwdriver-cd/data-schema/blob/master/models/build.js#L148
Have pasted the log to u in slack

tk3fftk:
Thanks for pointed and logs! I fixed.

minz1027:
screen shot 2017-03-03 at 11 03 23 am

@tk3fftk Cool! it works now! But the `meta` is under `/opt/sd`. Should we add a shortcut/link for it in the setup command? e.g. `ln -s /opt/sd/meta meta`

https://github.com/screwdriver-cd/launcher/blob/master/executor/executor.go#L100

tk3fftk:
@minz1027 Thank you!! I understood about path of `meta`. I commented about newline in issue.

PR #112, opened 2017-04-12
feat(300): install hab and sd-step in launcher

kumada626:
To use habitat in shared-step, this PR installs habitat in launcher. In addition, it installs sd-step go binary to wrap `hab` commands.
Detail is screwdriver-cd/screwdriver#300 .
This PR will be merged after https://github.com/screwdriver-cd/sd-step is built and released.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,6 +39,19 @@ RUN set -x \
    && rm tini-static.asc \
    && mv tini-static tini \
    && chmod +x tini \
+   # Install Habitat
+   && wget 'https://api.bintray.com/content/habitat/stable/linux/x86_64/hab-$latest-x86_64-linux.tar.gz?bt_package=hab-x86_64-linux' \
+   && tar -C . -ozxvf hab-\$latest-x86_64-linux.tar.gz?bt_package=hab-x86_64-linux \
+   && mkdir /opt/sd/bin \
+   && mv /opt/sd/hab-*-x86_64-linux/hab /opt/sd/bin/hab \
+   && chmod +x /opt/sd/bin/hab \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -39,6 +39,19 @@ RUN set -x \
...
+   && mkdir /opt/sd/bin \
+   && mv /opt/sd/hab-*-x86_64-linux/hab /opt/sd/bin/hab \
+   && chmod +x /opt/sd/bin/hab \
+   && rm -rf hab-* \
+   # Download sd-step
+   && wget -q -O - https://github.com/screwdriver-cd/sd-step/releases/latest \
+      | egrep -o '/screwdriver-cd/sd-step/releases/download/v[0-9.]*/sd-step' \
+      | wget --base=http://github.com/ -i - -O sd-step\
+   && mv /opt/sd/sd-step /opt/sd/bin/sd-step \
+   && chmod +x /opt/sd/bin/sd-step \
```
minz1027: Do we still need `sd-step` under `/opt/sd`? Why not rm it here?

kumada626: In discussion of screwdriver-cd/screwdriver#300, we planed first pass below.
>First pass:
> - Change launcher's Dockerfile
>    - install hab and it's wrapper command in Dockerfile
>    - install path: /opt/sd/bin/

Therefore, I put sd-step to `/opt/sd/bin`. However, other sd tools under `/opt/sd`. Should we put sd-step to `/opt/sd`?

minz1027: I feel it's more consistent that way :)

kumada626: OK, I'll fix it.

PR #113, opened 2017-04-12
fix(470): Add API URL environment variable

jerryzhang222:
This environment variable (`SD_API_URL`) will be read when publishing or validating a template in https://github.com/screwdriver-cd/template-main/pull/2

### DIFF
--- a/screwdriver/screwdriver.go
+++ b/screwdriver/screwdriver.go
@@ -42,6 +42,7 @@ type API interface {
 	UpdateStepStart(buildID int, stepName string) error
 	UpdateStepStop(buildID int, stepName string, exitCode int) error
 	SecretsForBuild(build Build) (Secrets, error)
+	GetAPIURL() (string, error)
 }
 
 // SDError is an error response from the Screwdriver API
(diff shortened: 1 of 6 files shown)


screwdriver/screwdriver.go
```diff
@@ -270,6 +271,10 @@ func (a api) put(url *url.URL, bodyType string, payload io.Reader) ([]byte, erro
 	return a.write(url, "PUT", bodyType, payload)
 }
 
+func (a api) GetAPIURL() string {
+	return a.baseURL
```
minz1027: This does not include namespace `v4`. You probably want to use `makeURL`

minz1027: Okay, seeing you guys have it here: https://github.com/screwdriver-cd/template-main/blob/46769f6ba2e3956c3d22c07cb5ad9a02d819fda9/index.js#L29.

But shall we just get it from launcher instead of hardcoding the v4 in different places?

jerryzhang222: I think the intention of having `template-main` check for the `SD_API_URL` (https://github.com/screwdriver-cd/template-main/blob/46769f6ba2e3956c3d22c07cb5ad9a02d819fda9/index.js#L25) was to make it flexible so if others deploy their own version of Screwdriver, it'll just use their API URL, which I think might not include a version.

Right now, `makeURL` seems to be hardcoding "v4" as the version in the URL.

jerryzhang222: Will anyone who deploys their own Screwdriver cluster also have the v4 at the end of their API URL?

d2lam: I feel like the `API URL` should just be the hostname. The namespace `v4` ties more to the endpoint. So if we move that endpoint to somewhere else, we just need to change the code in `template-main` instead of `launcher`. It makes more sense since the `template-main` interacts with the endpoint.

PR #114, opened 2017-04-18
fix: add /opt/sd to PATH instead of creating symlink

minz1027:
Add `/opt/sd` to PATH instead of creating symlink.
Because creating symlink requires root but it is not always the case.

Tested locally and works fine:
screen shot 2017-04-18 at 4 16 49 pm


d2lam:
Looks like it's working for in-a-box

PR #115, opened 2017-04-20
feat(473): Always run steps with the alwaysRun flag true

jerryzhang222:
Have the executor run always run steps that have the `alwaysRun` flag set as `true`.

Pull requests to define the `alwaysRun` flag in steps (all should be merged together): 

https://github.com/screwdriver-cd/models/pull/155
https://github.com/screwdriver-cd/build-bookend/pull/8
https://github.com/screwdriver-cd/config-parser/pull/34

Related: https://github.com/screwdriver-cd/screwdriver/issues/473

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -109,39 +109,52 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
 	f.Write([]byte(shargs))
 
 	cmds := build.Commands
+	var exitErr error
 
 	for _, cmd := range cmds {
-		if err := api.UpdateStepStart(buildID, cmd.Name); err != nil {
-			return fmt.Errorf("Updating step start %q: %v", cmd.Name, err)
-		}
+		if exitErr == nil || cmd.AlwaysRun {
+			if err := api.UpdateStepStart(buildID, cmd.Name); err != nil {
+				if exitErr == nil {
+					exitErr = fmt.Errorf("Updating step start %q: %v", cmd.Name, err)
+				}
+				continue
+			}
 
-		// Create step script file
-		stepFilePath := "/tmp/step.sh"
-		if err := createShFile(stepFilePath, cmd); err != nil {
-			return fmt.Errorf("Writing to step script file: %v", err)
-		}
+			// Create step script file
+			stepFilePath := "/tmp/step.sh"
+			if err := createShFile(stepFilePath, cmd); err != nil {
+				exitErr = fmt.Errorf("Writing to step script file: %v", err)
+				continue
+			}
 
-		// Generate guid for the step
-		guid := uuid.NewV4().String()
+			// Generate guid for the step
+			guid := uuid.NewV4().String()
 
-		// Set current running step in emitter
-		emitter.StartCmd(cmd)
-		fmt.Fprintf(emitter, "$ %s\n", cmd.Cmd)
+			// Set current running step in emitter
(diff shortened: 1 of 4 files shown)


executor/executor.go
```diff
@@ -109,39 +109,42 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
...
 
-		// Create step script file
-		stepFilePath := "/tmp/step.sh"
-		if err := createShFile(stepFilePath, cmd); err != nil {
-			return fmt.Errorf("Writing to step script file: %v", err)
-		}
+			// Create step script file
+			stepFilePath := "/tmp/step.sh"
+			if err := createShFile(stepFilePath, cmd); err != nil {
+				exitErr = fmt.Errorf("Writing to step script file: %v", err)
```
tkyi: If you take out the `return`s here, that means that even if there is an error when trying to `createShFile`, we will still continue on to create a guid and updateStepStop and everything

screwdriver/screwdriver.go
```diff
@@ -110,8 +110,9 @@ type Job struct {
 
 // CommandDef is the definition of a single executable command.
 type CommandDef struct {
-	Name string `json:"name"`
-	Cmd  string `json:"command"`
+	Name      string `json:"name"`
+	Cmd       string `json:"command"`
+	AlwaysRun string `json:"alwaysRun"`
```
d2lam: How come this is a string?

jerryzhang222: Does this part convert everything to a string before going into the launcher? https://github.com/screwdriver-cd/config-parser/blob/master/lib/phase/permutation.js#L38

executor/executor.go
jerryzhang222: I've added `continue` statements in each place where it would otherwise be returning. This will ensure the `alwaysRun` logic works. Does this make sense?

screwdriver/screwdriver.go
d2lam: I don't think so. The `permutations` is for the `matrix` field:
```
const matrix = Hoek.reach(doc, `jobs.${jobName}.matrix`, {
    default: {}
});
const permutations = keymbinatorial(matrix);
```

You can try running this:
https://gist.github.com/d2lam/f7b39753aa33b9ceac7327ed11a13243

It's showing that `alwaysRun` value is `false` as boolean

executor/executor.go
```diff
@@ -109,39 +109,46 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
...
+				exitErr = cmdErr
+				continue
+			}
 		}
 	}
 
 	f.Write([]byte{4}) // EOT
 
-	return nil
+	return exitErr
```
minz1027: `exitErr` is for the last failure step, this may cause problem when updating the whole build's status. 
Can we return the exitCode for the first failure step instead? I think that makes more sense.

minz1027: nvm, guess we don't care about the exit code for the whole build

d2lam:
Can't merge this yet due to: https://cd.screwdriver.cd/pipelines/1/builds/3615

We need to fix the schema for validator: https://github.com/screwdriver-cd/data-schema/blob/master/api/validator.js#L8-L12 to allow `alwaysRun`

We can just expose `SCHEMA_FULL_STEP_OBJECT` from `config/job` and call `job.full_step` there.

PR #119, opened 2017-08-25
	fix(#525): Move from bufio.Scanner to bufio.Reader

stjohnjohnson:
Fixed the problem of reading long-lines.

See [this](https://golang.org/pkg/bufio/#Scanner) for more info:  
> Programs that need more control over error handling or large tokens, 
> or must run sequential scans on a reader, should use bufio.Reader 
> instead.

I pulled the helper function from: https://stackoverflow.com/a/12206365


d2lam:
Related: https://github.com/screwdriver-cd/screwdriver/issues/525

PR #126, opened 2017-10-27
fix: can get prNumber even if job name is PR-1:main

kumada626:
## Context
Now, `PR-1:main` is allowed as job name in job model.
However, when job name is `PR-1:main`, SD_PULL_REQUEST is empty because prNumber's regexp doesn't match `PR-1:main`.

## Objective
Fix prNumber's regexp


petey:
good find.

PR #133, opened 2017-12-08
feat(473): always run teardown bookends

minz1027:
- Run setup bookends and user steps in a single shell as what it is now
- Exit the single shell and run teardown bookends in separate shells


executor/executor.go
```diff
@@ -134,40 +133,76 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
...
+				}
+				exportString := strings.Join(exportCmd, " && ")
+
+				exec.Command(exportString)
+
+				// set teardownFlag so this block won't run again
+				teardownFlag = true
+			}
+
+			exec.Command(cmd.Cmd)
```
minz1027: This whole part should be something similar to this:
https://github.com/screwdriver-cd/launcher/blob/7be81506e9b1e34a2cab336b85fd1509bd231363/executor/executor.go#L32
We can probably just wrap in inside a separate function.

executor/executor.go
```diff
@@ -134,40 +133,76 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
...
 
-		if cmdErr != nil {
-			return cmdErr
+		// Set first error flag
+		if (firstError == nil) && (cmdErr != nil) {
+			firstError = cmdErr
 		}
 	}
 
 	f.Write([]byte{4}) // EOT
```
minz1027: Should move this up to replace `exec.command('exit')`

executor/executor.go
d2lam: Cool! 👍

executor/executor.go
d2lam: Fixed

PR #138, opened 2018-01-09
feat: set store url to env

yuichi10:
## Context
Launcher set store url to job environment.

When we use sd-cmd, it will communicate with Store APIs. But now sd-cmd can not get Store url from job environment. Thus I make it possible to set store api in job environment by launcher.

## Objective
Allow to add new argument of `store-uri` and set Store url in job environment by using the argument. 

The value of key is `SD_STORE_URL`.

## References
[command document](https://github.com/screwdriver-cd/screwdriver/blob/master/design/commands.md)
[sd-cmd PR](https://github.com/screwdriver-cd/sd-cmd/pull/4)

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -161,7 +161,7 @@ func prNumber(jobName string) string {
 	return matched[1]
 }
 
-func launch(api screwdriver.API, buildID int, rootDir, emitterPath string, metaSpace string) error {
+func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, storeURL string) error {
 	emitter, err := newEmitter(emitterPath)
 	if err != nil {
 		return err
@@ -294,6 +294,7 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath string, metaS
 		"SD_API_URL":             apiURL,
 		"SD_BUILD_URL":           apiURL + "builds/" + strconv.Itoa(buildID),
 		"SD_BUILD_SHA":           b.SHA,
+		"SD_STORE_URL":           fmt.Sprintf("%s/%s/", storeURL, "v1"),
 	}
 
 	secrets, err := api.SecretsForBuild(b)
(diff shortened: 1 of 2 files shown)


yuichi10:
After merged this request, I will fix the [guide page of environment-variables](https://github.com/screwdriver-cd/guide/blob/master/docs/user-guide/environment-variables.md)

launch.go
```diff
@@ -440,6 +441,11 @@ func main() {
 			Usage: "Location of meta temporarily",
 			Value: "/sd/meta",
 		},
+		cli.StringFlag{
+			Name:  "store-uri",
```
yoshwata: Store-uri and Store-url coexist. Does it express important difference between them?

launch.go
```diff
@@ -448,6 +454,7 @@ func main() {
 		workspace := c.String("workspace")
 		emitterPath := c.String("emitter")
 		metaSpace := c.String("meta-space")
+		storeURL := c.String("store-uri")
```
yoshwata: Store-uri and Store-url coexist. Does it express important difference?

yuichi10: Screwdriver api is also coexist uri and url thus I wrote by the way as same as Screwdriver api.

https://github.com/screwdriver-cd/launcher/blob/master/launch.go#L446

yoshwata: Understood. Thank you.

launch.go
```diff
@@ -440,6 +441,11 @@ func main() {
 			Usage: "Location of meta temporarily",
 			Value: "/sd/meta",
 		},
+		cli.StringFlag{
```
d2lam: Does that mean we need to also make changes to all the executors in order for this change to work. 
For example:  https://github.com/screwdriver-cd/executor-k8s/blob/master/config/pod.yaml.tim#L27

yuichi10: Definitely. I will also send PR for these repositories.

d2lam: Cool! Thanks!

PR #148, opened 2018-02-07
fix: Moving SD_SHELL_BIN to args and adding debug info to start

stjohnjohnson:
I moved `SD_SHELL_BIN` from a hidden environment variable to an argument that can be passed on the command line.

Also fixed startup and teardown to use the shell that was specified.

Finally, added some extra output to the start of the build to know more about the running job.


habitat/plan.sh
```diff
@@ -18,6 +18,7 @@ scaffolding_go_build_deps=(
     gopkg.in/kr/pty.v1
     gopkg.in/myesui/uuid.v1
     gopkg.in/urfave/cli.v1
+	gopkg.in/fatih/color.v1
```
tkyi: Indentation looks a little off here

stjohnjohnson: Fixed

PR #149, opened 2018-02-09
feat(habitat): Putting all Habitat files in /hab

stjohnjohnson:
This stores everything in `/hab` since Habitat doesn't support alternative root directories.  After this, we can mount `/hab` into the build container and have access to pre-installed packages (like bash, git, and more!).  Additionally, teams running their own cluster can install their own packages as well.

_Note: increases size from 38mb to 91mb 😞 _

Related to: https://github.com/screwdriver-cd/screwdriver/issues/877

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 FROM alpine:3.4
 MAINTAINER The Screwdrivers <screwdriver.cd>
 
-# Download Launcher, Log Service, and Tini setup
 WORKDIR /opt/sd
 RUN set -x \
    # Alpine ships with musl instead of glibc (this fixes the symlink)
@@ -11,6 +10,7 @@ RUN set -x \
    && apk add --no-cache --update ca-certificates \
    && apk add --virtual .build-dependencies wget \
    && apk add --virtual .build-dependencies gpgme \
+
    # Download Launcher
    && wget -q -O - https://github.com/screwdriver-cd/launcher/releases/latest \
       | egrep -o '/screwdriver-cd/launcher/releases/download/v[0-9.]*/launcher_linux_amd64' \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -51,18 +52,26 @@ RUN set -x \
...
-   && chmod +x /opt/sd/bin/hab \
-   && rm -rf hab-* \
+   && mkdir -p /hab/bin /opt/sd/bin \
+   # Download Habitat Binary
+   && wget -O hab.tar.gz 'https://api.bintray.com/content/habitat/stable/linux/x86_64/hab-$latest-x86_64-linux.tar.gz?bt_package=hab-x86_64-linux' \
+   && tar -C . -ozxvf hab.tar.gz \
+   && mv hab-*/hab /hab/bin/hab \
+   && chmod +x /hab/bin/hab \
+   # @TODO Remove this, I don't think it belongs here.  We should use /hab/bin/hab instead.
+   && cp /hab/bin/hab /opt/sd/bin/hab \
```
catto: TODO comment +1
note: We need to fix [this](https://github.com/screwdriver-cd/sd-step/blob/master/sd-step.go#L26) once new launcher is released

Dockerfile
```diff
@@ -51,18 +52,26 @@ RUN set -x \
...
-   && mv /hab /opt/sd/hab-cache \
-   && ln -s /opt/sd/hab-cache /hab \
+   && /hab/bin/hab pkg install core/bash core/git \
+   # Cleanup Habitat Files
+   && rm -rf /hab/cache /opt/sd/hab.tar.gz /opt/sd/hab-* \
+   # Cleanup docs and man pages (how could this go wrong)
+   && find /hab -name doc -exec rm -r {} + \
+   && find /hab -name docs -exec rm -r {} + \
+   && find /hab -name man -exec rm -r {} + \
+   && rm -rf hab.tar.gz \
```
catto: Is `hab.tar.gz` already removed at L68?

PR #154, opened 2018-03-08
fix: Get meta from parent event when no parent build ID

tkyi:
## Context
When a user starts a detached job, it would be useful if they can use the meta from the entire parent event, rather than the parent build, since a detached job will not have a parent build ID.

## Objective
This PR uses meta from the parent event when no parent build ID exists.

*There will be more work in the future for builds with multiple parent build IDs to use the aggregate meta from their parent build IDs to rerun.*

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/904
Blocked by https://github.com/screwdriver-cd/data-schema/pull/209

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -192,6 +192,12 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 		return fmt.Errorf("Fetching Build ID %d: %v", buildID, err)
 	}
 
+	log.Printf("Fetching Event %d", b.EventID)
+	e, err := api.EventFromID(b.EventID)
+	if err != nil {
+		return fmt.Errorf("Fetching Event ID %d: %v", b.EventID, err)
+	}
+
 	log.Printf("Fetching Job %d", b.JobID)
 	j, err := api.JobFromID(b.JobID)
 	if err != nil {
@@ -204,8 +210,33 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 		return fmt.Errorf("Fetching Pipeline ID %d: %v", j.PipelineID, err)
 	}
 
+	// If no parent build ID, get the parent event meta
 	if b.ParentBuildID == 0 {
-		log.Printf("This build has no Parent Build, so fetching Meta is skipped")
+		log.Printf("Fetching Parent Event %d", e.ParentEventID)
+		pe, err := api.EventFromID(e.ParentEventID)
+		if err != nil {
+			return fmt.Errorf("Fetching Parent Event ID %d: %v", e.ParentEventID, err)
+		}
+
+		// Write to "meta.json" file
+		metaFile := "meta.json"
+
+		log.Printf("Creating Meta Space in %v", metaSpace)
+		err = createMetaSpace(metaSpace)
+		if err != nil {
+			return err
+		}
+
+		log.Printf("Fetching Parent Event Meta JSON %v", pe.ID)
+		peMetaByte, err := marshal(pe.Meta)
+		if err != nil {
+			return fmt.Errorf("Parsing Parent Event(%d) Meta JSON: %v", pe.ID, err)
+		}
+
+		err = writeFile(metaSpace+"/"+metaFile, peMetaByte, 0666)
+		if err != nil {
+			return fmt.Errorf("Writing Parent Event(%d) Meta JSON: %v", pe.ID, err)
+		}
 	} else {
 		log.Printf("Fetching Parent Build %d", b.ParentBuildID)
 		pb, err := api.BuildFromID(b.ParentBuildID)
--- a/launch.go
+++ b/launch.go
@@ -192,12 +192,6 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 		return fmt.Errorf("Fetching Build ID %d: %v", buildID, err)
 	}
 
-	log.Printf("Fetching Event %d", build.EventID)
-	event, err := api.EventFromID(build.EventID)
-	if err != nil {
-		return fmt.Errorf("Fetching Event ID %d: %v", build.EventID, err)
-	}
-
 	log.Printf("Fetching Job %d", build.JobID)
 	job, err := api.JobFromID(build.JobID)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


launch_test.go
```diff
@@ -833,6 +866,29 @@ func TestFetchParentBuildMeta(t *testing.T) {
...
+
+	api := mockAPI(t, TestEventID, TestJobID, 0, "RUNNING")
+	api.eventFromID = func(eventID int) (screwdriver.Event, error) {
+		if eventID == TestParentEventID {
+			return screwdriver.Event(FakeEvent{ID: TestParentEventID}), nil
+		}
+		return screwdriver.Event(FakeEvent{ID: TestEventID, ParentEventID: TestParentEventID}), nil
+	}
+	marshal = func(v interface{}) (result []byte, err error) {
+		return []byte("test"), fmt.Errorf("Testing parsing parent event meta")
```
FenrirUnbound: You should return `nil` if you don't care about the `result`

screwdriver/screwdriver.go
```diff
@@ -292,6 +300,21 @@ func (a api) BuildFromID(buildID int) (build Build, err error) {
...
 
+// EventFromID fetches and returns a Event object from its ID
+func (a api) EventFromID(eventID int) (event Event, err error) {
+	u, err := a.makeURL(fmt.Sprintf("events/%d", eventID))
+	body, err := a.get(u)
+	if err != nil {
+		return event, err
+	}
+
+	err = json.Unmarshal(body, &event)
```
FenrirUnbound: Why bother with returning an `Event` object when you're modifying what's passed in anyways?

launch.go
```diff
@@ -192,6 +192,12 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 		return fmt.Errorf("Fetching Build ID %d: %v", buildID, err)
 	}
 
+	log.Printf("Fetching Event %d", b.EventID)
+	e, err := api.EventFromID(b.EventID)
```
FenrirUnbound: Please use a more meaningful variable name than `e`

launch.go
```diff
@@ -204,8 +210,33 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 		return fmt.Errorf("Fetching Pipeline ID %d: %v", j.PipelineID, err)
 	}
 
+	// If no parent build ID, get the parent event meta
 	if b.ParentBuildID == 0 {
-		log.Printf("This build has no Parent Build, so fetching Meta is skipped")
+		log.Printf("Fetching Parent Event %d", e.ParentEventID)
+		pe, err := api.EventFromID(e.ParentEventID)
```
FenrirUnbound: Please use a more meaningful variable name than `pe`

launch_test.go
```diff
@@ -833,6 +866,29 @@ func TestFetchParentBuildMeta(t *testing.T) {
...
+	}
+
+	err := launch(screwdriver.API(api), TestEventID, TestWorkspace, TestEmitter, TestMetaSpace, TestStoreURL, TestShellBin)
+	want := "Parsing Parent Event(" + strconv.Itoa(TestParentEventID) + ") Meta JSON: Testing parsing parent event meta"
```
FenrirUnbound: Use `fmt` for formatting strings

launch.go
```diff
@@ -204,8 +210,33 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
...
+		if err != nil {
+			return fmt.Errorf("Fetching Parent Event ID %d: %v", e.ParentEventID, err)
+		}
+
+		// Write to "meta.json" file
+		metaFile := "meta.json"
```
d2lam: We can probably combine this and the code below in the `else`

launch.go
tkyi: I'm just staying consistent with other variables in the file/repo.
Having short 1-2 letter vars also seems to be a golang convention:
- http://wordaligned.org/articles/go-for-short-variable-names
- https://github.com/golang/go/wiki/CodeReviewComments#variable-names

If we change this one, we should change them all. Do you want me to change all of them?

FenrirUnbound: > If we change this one, we should change them all. Do you want me to change all of them?

Honestly, this code base started off with very limited Golang knowledge. We can either perpetuate to maintain the same low proficiency, or actually improve on the code bases we're working in.

launch_test.go
tkyi: Fixed

launch.go
```diff
@@ -204,8 +210,33 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
...
+			return err
+		}
+
+		log.Printf("Fetching Parent Event Meta JSON %v", pe.ID)
+		peMetaByte, err := marshal(pe.Meta)
+		if err != nil {
+			return fmt.Errorf("Parsing Parent Event(%d) Meta JSON: %v", pe.ID, err)
+		}
+
+		err = writeFile(metaSpace+"/"+metaFile, peMetaByte, 0666)
```
FenrirUnbound: `fmt`

launch.go
tkyi: Updated the variable names

launch.go
tkyi: What is the benefit in using `fmt` vs just doing a simple string concatenation here?

launch_test.go
tkyi: Fixed

launch.go
tkyi: Refactored code to try to combine some overlapping logic
Lmk if that looks okay

launch.go
```diff
@@ -187,72 +187,99 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
...
-	b, err := api.BuildFromID(buildID)
+	build, err := api.BuildFromID(buildID)
 	if err != nil {
 		return fmt.Errorf("Fetching Build ID %d: %v", buildID, err)
 	}
 
-	log.Printf("Fetching Job %d", b.JobID)
-	j, err := api.JobFromID(b.JobID)
+	log.Printf("Fetching Event %d", build.EventID)
+	event, err := api.EventFromID(build.EventID)
```
d2lam: We only need to make this call if there is no parent build

tkyi: Fixed 👍

PR #155, opened 2018-03-08
refactor: separate processing of user-defined steps and teardown steps

FenrirUnbound:
## Context

The current proposal for implementing a build timeout requires that tear-down commands always be executed when the build times out. This refactor splits the set of commands up as such.

## Objective

Split the user-defined steps and the tear-down steps. Process the user-defined commands first, then process the tear-down steps.

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -142,6 +142,23 @@ func doRunTeardownCommand(cmd screwdriver.CommandDef, emitter screwdriver.Emitte
 	return ExitOk, nil
 }
 
+func filterTeardowns(build screwdriver.Build) ([]screwdriver.CommandDef, []screwdriver.CommandDef) {
+	userCommands := []screwdriver.CommandDef{}
+	teardownCommands := []screwdriver.CommandDef{}
+
+	for _, cmd := range build.Commands {
+		isTeardown, _ := regexp.MatchString("sd-teardown-*", cmd.Name)
+
+		if isTeardown {
+			teardownCommands = append(teardownCommands, cmd)
+		} else {
+			userCommands = append(userCommands, cmd)
+		}
+	}
+
+	return userCommands, teardownCommands
+}
+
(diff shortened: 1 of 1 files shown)


d2lam:
Please DNM this yet. There is a bug in the previous launcher PR.

FenrirUnbound:
> Please DNM this yet. There is a bug in the previous launcher PR.

Why does that matter? The previous PR doesn't affect the executor module

d2lam:
@FenrirUnbound Because we wouldn't be able to pull this in. You will not be able to start any builds: https://beta.cd.screwdriver.cd/pipelines/93/builds/8302

FenrirUnbound:
> @FenrirUnbound Because we wouldn't be able to pull this in. You will not be able to start any builds: https://beta.cd.screwdriver.cd/pipelines/93/builds/8302

That's fine, but I fail t see how that affects this particular code change though. 

Whenever there's an infrastructure issue, are we changing _all_ PRs to DNM?

executor/executor.go
```diff
@@ -165,72 +182,67 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
...
-			if err := api.UpdateStepStop(buildID, cmd.Name, code); err != nil {
-				return fmt.Errorf("Updating step stop %q: %v", cmd.Name, err)
-			}
+		if err := api.UpdateStepStop(buildID, cmd.Name, code); err != nil {
+			return fmt.Errorf("Updating step stop %q: %v", cmd.Name, err)
 		}
 
-		// Set first error flag
-		if (firstError == nil) && (cmdErr != nil) {
+		if (firstError == nil) {
```
tkyi: Missing `&& (cmdErr != nil)` check. Do we no longer need it?

tkyi: Okay, that makes sense 👍

PR #157, opened 2018-03-14
feat: Merge parentBuildId metas

tkyi:
## Context
When a join occurs, we're currently just using the meta from the last build that triggered the join job to start. For example, if C requires A and B, if A finishes running first, B will trigger C to start after it is done. C would then be run with whatever metadata B had.
```
A 
     -->  C
B
```
We would like to merge all the metadata from all parent builds so that C can run using metadata from both A and B.

## Objective
This PR checks for multiple parent build IDs and merges the metadata for all of them.

## Related Links
Related to https://github.com/screwdriver-cd/screwdriver/issues/904
Blocked by https://github.com/screwdriver-cd/data-schema/pull/210

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -211,16 +211,37 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 	}
 
 	metaByte := []byte("")
-	parentID := 0
 	metaFile := "meta.json" // Write to "meta.json" file
-	metaSource := ""
+	metaLog := ""
 
 	// If no parent build ID, and no parent event, skip fetch meta
 	if build.ParentBuildID == 0 && event.ParentEventID == 0 {
 		log.Printf("This build has no Parent Build and no Parent Event, so fetching Meta is skipped")
 	} else {
-		// If has parent build, fetch from parent
-		if build.ParentBuildID != 0 {
+		if build.ParentBuildIDs != nil && len(build.ParentBuildIDs) > 0 { // If has multiple parent build IDs, merge their metadata
+			mergedMeta := make(map[string]interface{})
+
+			// Get meta from all parent builds
+			for _, pbID := range build.ParentBuildIDs {
+				pb, err := api.BuildFromID(pbID)
+				if err != nil {
+					return fmt.Errorf("Fetching Parent Build ID %d: %v", pbID, err)
+				}
+
+				// Save meta to mergedMeta object
+				for key, value := range pb.Meta {
+					mergedMeta[key] = value
+				}
+			}
+
+			log.Printf("Marshalling Merged Meta JSON %v", build.ParentBuildIDs)
+			metaByte, err = marshal(mergedMeta)
+			if err != nil {
+				return fmt.Errorf("Parsing Parent Builds(%v) Meta JSON: %v", build.ParentBuildIDs, err)
+			}
+
+			metaLog = fmt.Sprintf(`Builds(%v)`, build.ParentBuildIDs)
+		} else if build.ParentBuildID != 0 { // If has parent build, fetch from parent build
 			log.Printf("Fetching Parent Build %d", build.ParentBuildID)
 			parentBuild, err := api.BuildFromID(build.ParentBuildID)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


launch_test.go
```diff
@@ -958,6 +986,74 @@ func TestFetchParentBuildMetaWriteError(t *testing.T) {
...
+	}
+	if actual["batman"] != "robin" {
+		t.Errorf("Error is wrong, got '%v', expected '%v'", actual["batman"], "robin")
+	}
+	if actual["wonder"] != "woman" {
+		t.Errorf("Error is wrong, got '%v', expected '%v'", actual["wonder"], "woman")
+	}
+}
+
+func TestMergedMetaFromParentBuildsIDs(t *testing.T) {
```
minz1027: what is this test case trying to test? should be covered by `TestFetchParentBuildsMeta ` already, right?

tkyi: Removed

launch.go
```diff
@@ -213,20 +232,43 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
...
 	// If no parent build ID, and no parent event, skip fetch meta
-	if build.ParentBuildID == 0 && event.ParentEventID == 0 {
+	if len(parentBuildIDs) == 0 && event.ParentEventID == 0 {
 		log.Printf("This build has no Parent Build and no Parent Event, so fetching Meta is skipped")
 	} else {
-		// If has parent build, fetch from parent
-		if build.ParentBuildID != 0 {
-			log.Printf("Fetching Parent Build %d", build.ParentBuildID)
-			parentBuild, err := api.BuildFromID(build.ParentBuildID)
+		if len(parentBuildIDs) > 1 { // If has multiple parent build IDs, merge their metadata
```
FenrirUnbound: We could also make this an `else if`

launch.go
```diff
@@ -213,20 +232,43 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 	}
 
 	metaByte := []byte("")
-	parentID := 0
 	metaFile := "meta.json" // Write to "meta.json" file
-	metaSource := ""
+	metaLog := ""
+
+	parentBuildIDs := convertToArray(build.ParentBuildID)
```
d2lam: just a concern here: if there is no `build.ParentBuildID` then it **should** return `[]`?

launch_test.go
```diff
@@ -33,6 +33,8 @@ const (
 	TestSHA    = "abc123"
 )
 
+var TestParentBuildIDFloat interface{} = float64(1111)
```
FenrirUnbound: If it's not going to be updated, use a `const` instead of `var`

launch_test.go
```diff
@@ -865,6 +867,32 @@ func TestFetchParentBuildMeta(t *testing.T) {
 	}
 }
 
+func TestFetchParentBuildsMetaParseError(t *testing.T) {
+	TestParentBuildIDs := []float64{1111, 2222}
+	IDs := make([]interface{}, len(TestParentBuildIDs))
```
FenrirUnbound: Why can't we just initialize `IDs` with the test values?

launch.go
tkyi: Yes, I tried running it in the Golang playground and it looks like `convertToArray` should handle `nil` properly

launch.go
tkyi: We would have to duplicate the `create meta space section` or make a function out of it

launch.go
FenrirUnbound: I guess that depends on how Golang would unmarshal it if it's `0`

FenrirUnbound: > Yes, I tried running it in the Golang playground and it looks like convertToArray should handle nil properly

What if it's `undefined`?

launch_test.go
tkyi: I can't seem to initialize `IDs` in a simpler way. I think interface arrays are treated differently than int arrays.

launch_test.go
tkyi: I can't seem to assign that var as a `const`. I'm getting:
```
./launch_test.go:38: const initializer []float64 literal is not a constant
```

FenrirUnbound: Yeah, I guess it's because slices aren't immutable

FenrirUnbound: It's fine as is 😺

launch_test.go
FenrirUnbound: Fine as-is

tkyi: Moved the vars to a more global level

tkyi:
Darren's PR broke my tests :[

PR #158, opened 2018-03-14
feat: unmodifiable timeout

FenrirUnbound:
## Context

This is an initial proposal for killing a build after a period of time so they're not eating up system resources.

Further revisions will include:
* The ability to change the timeout limit
* Enforce a range of 15 minutes - 10 hours

## Objective

Use a non-modifiable build timeout for killing builds that take longer than 90 minutes.

## References 

* [Build Timeout issue](screwdriver-cd/screwdriver#545)

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -2,6 +2,7 @@ package executor
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -11,6 +12,7 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/screwdriver-cd/launcher/screwdriver"
 	"gopkg.in/kr/pty.v1"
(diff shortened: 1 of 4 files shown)


executor/executor.go
```diff
@@ -205,20 +214,35 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
...
+		completed <- cmdErr
 
-		if err := api.UpdateStepStop(buildID, cmd.Name, code); err != nil {
-			return fmt.Errorf("Updating step stop %q: %v", cmd.Name, err)
+		select {
+		case <- completed:
+			if (firstError == nil) {
+				firstError = cmdErr
+			}
+		case buildTimeout := <- invokeTimeout:
```
d2lam: This is only for userCommands? What happens if teardown step is the one that takes a long time? We should still stop the build right? Can we refactor to use this for both.

jithin1987: Considering post-build time (uploading artifacts) is influenced by the number of files generated in actual build operations it should be covered by the time out. In fact timeout should cover all, pre build, build steps, and post build

FenrirUnbound: > Considering post-build time (uploading artifacts) is influenced by the number of files generated in actual build operations it should be covered by the time out. In fact timeout should cover all, pre build, build steps, and post build, there is no need for separate settings for each steps.

That opens up the possibility that when a test times out, there are no artifacts saved. This would make debugging a hanging build very difficult if the artifacts are the only way to investigate the issue.

jithin1987: That's a valid concern, lets start with timeout covering pre & user build steps, then we can introduce a separate one for post builds steps.

FenrirUnbound: I still believe Dao's concern is legitimate. I'll add it to the feature issue (https://github.com/screwdriver-cd/screwdriver/issues/545) and we can create a discussion around that piece or at least track the decisions there.

FenrirUnbound: Added it: https://github.com/screwdriver-cd/screwdriver/issues/545#issuecomment-373210885

PR #160, opened 2018-03-15
feat: allow build timeout to be configured

FenrirUnbound:
## Context

This extends the initial proposal for killing a build after a period of time so they're not eating up system resources.

Further revisions will include:
* Enforce a range of 15 minutes - 10 hours

## Objective

Allow the build timeout for killing builds to be configurable. When not given an explicit value (whether through CLI flag or environment variable), use the default 90 minutes.

## References 

* [Build Timeout issue](https://github.com/screwdriver-cd/screwdriver#545)

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -171,7 +171,7 @@ func prNumber(jobName string) string {
 	return matched[1]
 }
 
-func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, storeURL, shellBin string) error {
+func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, storeURL, shellBin string, buildTimeout int) error {
 	emitter, err := newEmitter(emitterPath)
 	if err != nil {
 		return err
@@ -354,7 +354,7 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 		return fmt.Errorf("Updating sd-setup-launcher stop: %v", err)
 	}
 
-	return executorRun(w.Src, env, emitter, build, api, buildID, shellBin, DefaultTimeout)
+	return executorRun(w.Src, env, emitter, build, api, buildID, shellBin, buildTimeout)
 }
 
 func createEnvironment(base map[string]string, secrets screwdriver.Secrets, build screwdriver.Build) []string {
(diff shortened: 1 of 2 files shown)


launch.go
```diff
@@ -507,24 +513,24 @@ func main() {
...
 		if err != nil {
-			log.Printf("Error creating Screwdriver API %v: %v", buildID, err)
-			exit(screwdriver.Failure, buildID, nil, metaSpace)
+			return cli.ShowAppHelp(c)
 		}
 
-		defer recoverPanic(buildID, api, metaSpace)
-
-		launchAction(api, buildID, workspace, emitterPath, metaSpace, storeURL, shellBin)
+		fmt.Printf("%v\n", url)
```
tkyi: Did you mean to leave all these print statements here?

FenrirUnbound: Good catch! Amended.

FenrirUnbound:
> Go is not particularly my strong suit, but it looks like there is a test missing that determines that a failure occurred because a build exceeded a timeout. I don't think the launcher really needs to enforce the timeout range, that can happen when we validate a user's config. So there could be a test where the timeout is 1 second, and it runs a step that is `sleep(2)`.

In the main launcher module, we don't test the flags being passed in. Since the tests are mainly functional tests, we're only asserting the interfaces and that they're being used correctly.

In a previous PR, we introduced a test for the actual timeout: https://github.com/screwdriver-cd/launcher/pull/158/files#diff-6d9a596d0d155b4342a598ba359efc8bR274. This test in particular doesn't do many assertions other than the timeout happened. In a future PR, we have a cherry-picked commit that asserts which individual steps were executed so it's more explicit on which steps were supposed to execute and which were supposed to time out.

launch.go
```diff
@@ -496,6 +496,12 @@ func main() {
 			Value:  "/bin/sh",
 			EnvVar: "SD_SHELL_BIN",
 		},
+		cli.IntFlag{
+			Name:   "build-timeout",
+			Usage:  "Maximum number of seconds to allow a build to run",
+			Value:  DefaultTimeout,
+			EnvVar: "SD_BUILD_TIMEOUT",
```
minz1027: If it is a `flag` or `ENV`, how we are going to pass it in? are we going to change all executor to have a new ENV `SD_BUILD_TIMEOUT `? https://github.com/screwdriver-cd/executor-k8s/blob/master/config/pod.yaml.tim

FenrirUnbound: If a flag isn't set and the environment variable isn't set, the default timeout of 90 minutes will be used

minz1027: But we want to support configurable defaultTimeout and also user defined timeout right? The current implementation just make that a bit hard to achieve

FenrirUnbound: How do we currently pass in all the variables to the launcher (e.g., build ID, emitter path, etc)?

minz1027: moved it to the issue 👍

PR #161, opened 2018-03-16
fix: defer command execution to subprocess

FenrirUnbound:
## Context

The blocking operation was masked by the `MockEmitter` stub. By forcing it to block the current process with a `time.Sleep`, we were able to replicate how the timeout wasn't being executed.

## Objective

Execute the `doRunCommand` in a separate go routine.

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -216,7 +216,8 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
 		// Generate guid for the step
 		guid := uuid.NewV4().String()
 
-		completed := make(chan error, 1)
+		runErr := make(chan error, 1)
+		eCode := make(chan int, 1)
 
 		// Set current running step in emitter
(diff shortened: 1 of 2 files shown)


executor/executor.go
```diff
@@ -216,22 +216,27 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
...
 		emitter.StartCmd(cmd)
 		fmt.Fprintf(emitter, "$ %s\n", cmd.Cmd)
 
 		fReader := bufio.NewReader(f)
 
-		code, cmdErr = doRunCommand(guid, stepFilePath, emitter, f, fReader)
-		completed <- cmdErr
+		go func() {
+			runCode, rcErr := doRunCommand(guid, stepFilePath, emitter, f, fReader)
+			eCode <- runCode
```
tkyi: adding an `// exit code` comment would be nice here

PR #174, opened 2018-05-02
fix: Copy metadata to artifacts

tkyi:
## Context
Some users would like to know what's in their metadata at the end of each build (without having to print each one out).

## Objective
This PR copies the metadata to the artifacts folder at the end of the build.

## Related links
There's no built-in way to copy files in golang, so I used this page for reference: https://shapeshed.com/copy-a-file-in-go/

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -115,10 +115,31 @@ func doRunCommand(guid, path string, emitter screwdriver.Emitter, f *os.File, fR
 	return copyLinesUntil(fReader, emitter, guid)
 }
 
+// Copy the src file to dst.
+var copyFiles = func(src, dst string) error {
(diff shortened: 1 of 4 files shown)


jithin1987:
We should we put user meta information in `ARTIFACTS_DIR` and have it persisted in S3 for others to see ? Especially when we don't control nature of information user puts in meta, why make it available explicitly ?

I will be more comfortable if this is exposed as a shared command (`print-meta`) which user can call explicitly rather than building this logic inside SD, where users have no way to opt-out.

tkyi:
After discussing, we've decided exposing an environment variable for where the metadata lives makes more sense so the user can choose whether to cat the meta file or not.

PR #177, opened 2018-06-04
fix: defer emitter creation to container start-up

FenrirUnbound:
## Context

Docker 1.18 is unsupported due to a dependency's-dependency update in `containerd/continuity` where it fails to handle fifos files (which the `emitter` is).

## Objective 

Create the FIFO emitter at container start-up instead of creating it during the Docker image compilation. This should address issues with Docker 1.18 being unable to handle fifos files

Fixes screwdriver-cd/screwdriver#979

## References

* screwdriver-cd/screwdriver#979

### DIFF
--- a/Docker/launcher_entrypoint.sh
+++ b/Docker/launcher_entrypoint.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+# Create FIFO for emitter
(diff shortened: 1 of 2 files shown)


petey:
Docker 1.17 is a non-starter on my mac, I'm hit with a bug that 18.03 fixes `Fix Docker for Mac not starting due to socket file paths being too long`

PR #184, opened 2018-06-14
fix: allow to use API without coverage plugin

tk3fftk:
Launcher failed to launch a build by getting coverage info failure with API which doesn't have coverage plugin.  
This PR allows to use API without coverage plugin.

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -387,7 +387,7 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 	// Add coverage env vars
 	coverageInfo, err := api.GetCoverageInfo()
 	if err != nil {
-		return fmt.Errorf("Getting coverage info for build %v", build.ID)
+		log.Printf("Failed to get coverage info for build %v so skip it\n", build.ID)
 	}
 
 	for key, value := range coverageInfo.EnvVars {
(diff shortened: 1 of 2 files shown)


launch.go
```diff
@@ -387,7 +387,7 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 	// Add coverage env vars
 	coverageInfo, err := api.GetCoverageInfo()
 	if err != nil {
-		return fmt.Errorf("Getting coverage info for build %v", build.ID)
+		log.Printf("Failed to get coverage info for build %v so skip it\n", build.ID)
 	}
 
 	for key, value := range coverageInfo.EnvVars {
```
wdstar: We should not execute this `for` loop, if an error occurs.

tk3fftk: fixed!

catto: coverageInfo will be empty map object according to the type of the func, so this works as a zero-length loop.
https://github.com/screwdriver-cd/launcher/blob/master/screwdriver/screwdriver.go#L298

tk3fftk: yes, I think it's tested by added test case. But not executing is looks better.

tkyi:
Thank you!

PR #187, opened 2018-06-20
[DNM] fix: allow blocked status

d2lam:
Needs this so that `BLOCKED` can proceed to `RUNNING`
Related: https://github.com/screwdriver-cd/screwdriver/issues/653


tkyi:
Do we also need to add UNSTABLE?

d2lam:
@tkyi  I don't think it is possible. It cannot go to `UNSTABLE` unless the build is already at `RUNNING`. However, there is no harm to add it in case some logic changes later. I can add it.

PR #190, opened 2018-06-25
feat: support user-teardown

d2lam:
Allow user-defined teardown steps. They will be run in separate shells, just like the screwdriver teardown steps. 

_**Will look into passing environment variables in a second pass.**_ 

Related: https://github.com/screwdriver-cd/screwdriver/issues/1126

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -189,8 +189,8 @@ func filterTeardowns(build screwdriver.Build) ([]screwdriver.CommandDef, []screw
 	userTeardownCommands := []screwdriver.CommandDef{}
 
 	for _, cmd := range build.Commands {
-		isSdTeardown, _ := regexp.MatchString("sd-teardown-*", cmd.Name)
-		isUserTeardown, _ := regexp.MatchString("teardown-*", cmd.Name)
+		isSdTeardown, _ := regexp.MatchString("^sd-teardown-*", cmd.Name)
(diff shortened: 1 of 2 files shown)


executor/executor.go
```diff
@@ -183,21 +183,25 @@ func handleBuildTimeout(f *os.File, timeoutErr error) {
...
+func filterTeardowns(build screwdriver.Build) ([]screwdriver.CommandDef, []screwdriver.CommandDef, []screwdriver.CommandDef) {
 	userCommands := []screwdriver.CommandDef{}
-	teardownCommands := []screwdriver.CommandDef{}
+	sdTeardownCommands := []screwdriver.CommandDef{}
+	userTeardownCommands := []screwdriver.CommandDef{}
 
 	for _, cmd := range build.Commands {
-		isTeardown, _ := regexp.MatchString("sd-teardown-*", cmd.Name)
+		isSdTeardown, _ := regexp.MatchString("sd-teardown-*", cmd.Name)
+		isUserTeardown, _ := regexp.MatchString("teardown-*", cmd.Name)
```
tkyi: I think this regexp will overlap with `sd-teardown-*`, since the MatchString 
You need to use `^sd-teardown-*` and `^teardown-*` to match the beginning of the string

d2lam: Thank you. I'll fix it

PR #193, opened 2018-07-04
fix: [small] fix teardown regexp

tk3fftk:
Just `teardown` step should not be added to user defined teardown. It may work unintentionally if users already use `teardown` step name.
This PR fixes regexp not to include just `teardown` step.
It also does some refactor.

### DIFF
--- a/executor/.#executor.go
+++ b/executor/.#executor.go
@@ -1,0 +1,1 @@
+hitakats@MBP13JAQ-A039.local.6385
--- a/executor/.#executor.go
+++ b/executor/.#executor.go
@@ -1,1 +1,0 @@
-hitakats@MBP13JAQ-A039.local.6385
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -117,7 +117,7 @@ func doRunCommand(guid, path string, emitter screwdriver.Emitter, f *os.File, fR
 // Executes teardown commands
 func doRunTeardownCommand(cmd screwdriver.CommandDef, emitter screwdriver.Emitter, env []string, path, shellBin string) (int, error) {
 	shargs := []string{"-e", "-c"}
-	shargs = append(shargs, "export PATH=$PATH:/opt/sd && "+cmd.Cmd)
+	shargs = append(shargs, "export PATH=$PATH:/opt/sd && " + cmd.Cmd)
 	c := exec.Command(shellBin, shargs...)
 
 	emitter.StartCmd(cmd)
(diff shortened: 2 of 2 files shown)


executor/executor.go
```diff
@@ -189,8 +188,8 @@ func filterTeardowns(build screwdriver.Build) ([]screwdriver.CommandDef, []screw
 	userTeardownCommands := []screwdriver.CommandDef{}
 
 	for _, cmd := range build.Commands {
-		isSdTeardown, _ := regexp.MatchString("^sd-teardown-*", cmd.Name)
-		isUserTeardown, _ := regexp.MatchString("^teardown-*", cmd.Name)
+		isSdTeardown, _ := regexp.MatchString("^sd-teardown-.*", cmd.Name)
```
jithin1987: can we change these to `^sd-teardown-.+`

executor/executor.go
```diff
@@ -118,7 +117,7 @@ func doRunCommand(guid, path string, emitter screwdriver.Emitter, f *os.File, fR
 // Executes teardown commands
 func doRunTeardownCommand(cmd screwdriver.CommandDef, emitter screwdriver.Emitter, env []string, path, shellBin string) (int, error) {
 	shargs := []string{"-e", "-c"}
-	shargs = append(shargs, "export PATH=$PATH:/opt/sd && " + cmd.Cmd)
+	shargs = append(shargs, "export PATH=$PATH:/opt/sd && "+cmd.Cmd)
```
jithin1987: seems like we have  mix of styling around string concatenation operator.

executor/executor.go
d2lam: Makes sense. so that `teardown-` shouldn't match?

jithin1987: yes

executor/executor.go
tk3fftk: I think we should unify the writing. Should we choose no space or with space?

executor/executor.go
tk3fftk: fixed!

executor/.#executor.go
```diff
@@ -0,0 +1 @@
+hitakats@MBP13JAQ-A039.local.6385
```
tkyi: Probably don't want to commit this file?

tk3fftk: sorry, deleted!

tk3fftk:
@tkyi @jithin1987 deleted unused file 😄

PR #196, opened 2018-07-13
fix: pull in new log-service

d2lam:


tkyi:
Does this work without any code changes?

PR #198, opened 2018-07-19
Add core/zip from habitat

pranavrc:
# Context

https://github.com/screwdriver-cd/artifact-bookend/pull/23

To be able to zip artifact files before uploading to store, we need `zip` in all the images

# Objectives

- [x] Install `zip` from habitat

# Related links

PR to zip artifacts before uploading - https://github.com/screwdriver-cd/artifact-bookend/pull/23


d2lam:
Do we need to add to here to keep the habitat way of building work?
https://github.com/screwdriver-cd/launcher/blob/master/habitat/plan.sh#L9-L11

PR #200, opened 2018-07-25
fix: Pass env to teardown

d2lam:
Passing environment variables to teardown steps
In `trap`, we do `env` to a file and do proper parsing:
- escape double quote
- add double quotes around the value
In beginning of each teardown, we source the file

Related: https://github.com/screwdriver-cd/screwdriver/issues/1192

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -243,7 +243,7 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
 		// trap EXIT, echo the last step ID and write ENV to /tmp/buildEnv
 		"finish() { " +
 		"EXITCODE=$?; " +
-		exportEnvCmd + " && " +
+		exportEnvCmd + "; " +
 		"echo $SD_STEP_ID $EXITCODE; }",    //mv newfile to file
 		"trap finish EXIT;\n",
 	}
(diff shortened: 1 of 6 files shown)


executor/executor.go
```diff
@@ -215,13 +221,33 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
...
+
 	// Run setup commands
 	setupCommands := []string{
 		"set -e",
 		"export PATH=$PATH:/opt/sd",
-		"finish() { echo $SD_STEP_ID $?; }",
+		// trap EXIT, echo the last step ID and write ENV to /tmp/buildEnv
+		"finish() { " +
+		"EXITCODE=$?; " +
+		exportEnvCmd + " && " +
```
minz1027: If the exportEnvCmd failed, then this whole cmd will break here and won’t be able to print out the step exit code?

executor/executor.go
```diff
@@ -294,6 +320,13 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
 	}
 
 	teardownCommands := append(userTeardownCommands, sdTeardownCommands...)
+	cleanup := false
+
+	// Previous user steps ran successfully and there is no teardown step to run, clean up
+	if len(teardownCommands) == 0 && firstError == nil {
```
minz1027: Is this only for unit test? And now unit tests write to seperate file, we probably won’t need this logic any more

d2lam: No, this logic is for cleaning up for cases where all steps passed and there is no teardown step. For example:
```
- ls 
- echo hello
```
In this case, it will not go to `doRunTeardown` and no cleanup will be called. So we need this here for cleaning up.

minz1027: kk got it

PR #203, opened 2018-08-01
feat(1167): [4][small] Add fetch flag

kumada626:
## Context

To fetch build token in actual build execution, we fixed to exchange temporal token to build token in launcher process. 
However, also log-service uses token which is passed by `executor-*`.
To pass fetched build token to log-service and launcher, add the flag to launcher which only fetch token in launcher process, and use this flag to set fetched token before launcher and log-service processes start.
Detail: https://github.com/screwdriver-cd/screwdriver/issues/1167#issuecomment-409441739

## Objective

Add flag to only fetch build token in launcher process.

Add wrapper script to run build in multiple executors.

## References

issue: https://github.com/screwdriver-cd/screwdriver/issues/1167
issue comment: https://github.com/screwdriver-cd/screwdriver/issues/1167#issuecomment-409441739

PR:
executor-k8s: https://github.com/screwdriver-cd/executor-k8s/pull/89
hyperctl-image: https://github.com/screwdriver-cd/hyperctl-image/pull/24
executor-docker: https://github.com/screwdriver-cd/executor-docker/pull/28
executor-jenkins: https://github.com/screwdriver-cd/executor-jenkins/pull/24

### DIFF
--- a/Docker/run.sh
+++ b/Docker/run.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 # wrapper script for run build in multiple executors.
-SD_TOKEN=`/opt/sd/launch --only-fetch-token --token "$1" --api-uri $2 --store-uri $3 --emitter /sd/emitter --build-timeout $4 $5` && (/opt/sd/launch --api-uri $2 --store-uri $3 --emitter /sd/emitter --build-timeout $4 $5 & /opt/sd/logservice --emitter /sd/emitter --api-uri $2 --store-uri $3 --build $5 & wait $(jobs -p))
+SD_TOKEN=`/opt/sd/launch --only-fetch-token --token "$1" --api-uri "$2" --store-uri "$3" --emitter /sd/emitter --build-timeout "$4" "$5"` && (/opt/sd/launch --api-uri "$2" --store-uri "$3" --emitter /sd/emitter --build-timeout "$4" "$5" & /opt/sd/logservice --emitter /sd/emitter --api-uri "$2" --store-uri "$3" --build "$5" & wait $(jobs -p))
--- a/Docker/run.sh
+++ b/Docker/run.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+
+# wrapper script for run build in multiple executors.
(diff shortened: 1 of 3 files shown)


launch.go
```diff
@@ -562,6 +562,10 @@ func main() {
 			Value:  DefaultTimeout,
 			EnvVar: "SD_BUILD_TIMEOUT",
 		},
+		cli.BoolFlag{
+			Name:  "fetch-flag",
```
wdstar: We should use easily understandable flag name.
How about `fetch-token-only` or `only-fetch-token` ?

kumada626: It makes sense. Fix flag name to `only-fetch-token`.

launch.go
```diff
@@ -587,9 +592,16 @@ func main() {
 
 		buildToken, err := temporalApi.GetBuildToken(buildID, c.Int("build-timeout"))
 		if err != nil {
+			// If token is build scope token, GetBuildToken returns 403 error but need not exit
 			log.Printf("Error getting Build Token %v: %v", buildID, err)
 		}
 
+		if fetchFlag {
```
jithin1987: we shouldn't be fetching `buildToken` if `fetchFlag` is not passed.

kumada626: I changed the scope of fetchFlag if statement to only fetch `buildToken` if `fetchFlag` is passed.

Docker/run.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+# wrapper script for run build in multiple executors.
+SD_TOKEN=`/opt/sd/launch --only-fetch-token --token "$1" --api-uri $2 --store-uri $3 --emitter /sd/emitter --build-timeout $4 $5` && (/opt/sd/launch --api-uri $2 --store-uri $3 --emitter /sd/emitter --build-timeout $4 $5 & /opt/sd/logservice --emitter /sd/emitter --api-uri $2 --store-uri $3 --build $5 & wait $(jobs -p))
```
jithin1987: why not `"` all variables ?

kumada626: There is no specific reason not to `"` all variables. I fixed to `"` all variables.

PR #206, opened 2018-08-09
fix: write nops1 to a different file

d2lam:
It's possible that during this part, `$newfile` exists and the check for it in `doRunTeardown` passes and tries to source the env file. 

```
sed '/^PS1=.*/d' $file > $newfile &&
mv $newfile $file
```

This PR uses a different filename and make $newfile only available at the end

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -247,6 +247,20 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
 		"trap finish EXIT;\n",
 	}
 
+	 // The above script does the following
+	 // export PATH=$PATH:/opt/sd &&
+	 // finish() {
+	 //   EXITCODE=$?;
+	 //   prefix='export '; file=/tmp/env; nops1=/tmp/env_nops1; newfile=/tmp/env_export; env > $file &&
+	 //   sed '/^PS1=.* /d' $file > $nops1 &&
+	 //   while read -r line; do
+	 //   escapeQuote=`echo $line | sed 's/"/\\\"/g'` &&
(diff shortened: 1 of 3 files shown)


d2lam:
Cannot add the `|| echo 'Failed to get ENV'` because in some shell (mine is `zsh`) it will give error:
```
echo hi && (. /tmp/testTeardownEnv_export || echo 'Failed to get environment variables' ) && echo bye
hi
zsh(82751,0x7fff98ad2380) malloc: *** error for object 0x7fa203c1e680: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
[1]    82751 abort      ( . /tmp/testTeardownEnv_export || echo 'Failed to get environment variables')
```

For `sh` and `bash`, it is ok.

executor/executor.go
```diff
@@ -241,13 +243,29 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
...
+	 // export PATH=$PATH:/opt/sd &&
+	 // finish() {
+	 //   EXITCODE=$?;
+	 //   prefix='export '; file=/tmp/env; nops1=/tmp/env_nops1; newfile=/tmp/env_export; env > $file &&
+	 //   sed '/^PS1=.* /d' $file > $nops1 &&
+	 //   while read -r line; do
+	 //   escapeQuote=`echo $line | sed 's/"/\\\"/g'` &&
+	 //   newline=`echo $escapeQuote | sed 's/\([A-Za-z_][A-Za-z0-9_]*\)=\(.*\)/\1="\2"/'` &&
+	 //   echo ${prefix}$newline;
+	 //   done < $nops1 > $newfile;
```
minz1027: Need to update this comment

PR #209, opened 2018-08-14
fix: escape new line

d2lam:
Currently, if ENV or SECRET has newline character, the sourcing env file will fail. 
This PR escape newline character


d2lam:
Ah this might have some issue, for removing the last `\n`
In `sh` needs to do this `sed 's/..$/\\n/'`
While in `bash` needs to do this `sed 's/..$//'`

d2lam:
@jithin1987  👍  Please review here: https://github.com/screwdriver-cd/launcher/pull/210

PR #216, opened 2018-09-07
fix(979): change the emitter path

minz1027:
cannot mkfifo `/opt/sd/emitter` for k8s-vm since the share mount /opt/sd is `read-only`

https://git.ouroath.com/cocktails-screwdriver/hyperctl-docker/blob/master/scripts/hyper-pod-template.json#L23

### DIFF
--- a/Docker/launcher_entrypoint.sh
+++ b/Docker/launcher_entrypoint.sh
@@ -16,7 +16,7 @@ smart_run () {
 
 # Create FIFO for emitter
 # https://github.com/screwdriver-cd/screwdriver/issues/979
-smart_run mkfifo -m 666 /opt/sd/emitter
+smart_run mkfifo -m 666 /sd/emitter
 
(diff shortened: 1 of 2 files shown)


Docker/launcher_entrypoint.sh
```diff
@@ -16,7 +16,8 @@ smart_run () {
 
 # Create FIFO for emitter
 # https://github.com/screwdriver-cd/screwdriver/issues/979
-smart_run mkfifo -m 666 /opt/sd/emitter
+smart_run mkdir -p /sd
+smart_run mkfifo -m 666 /sd/emitter
```
jithin1987: is the only dynamically created file in the whole ecosystem which needs to be written into ?

minz1027: yes, other files are mounted in `/opt/sd`

PR #228, opened 2018-10-24
exec tini instead of subshell tini

lgfausak:
I think an _exec_ for tini would be better here.  It used to be that tini was the entrypoint, and thus it was PID 1.  Now it is being run as a subshell from the new entrypoint launcher_entrypoint.sh.  This change causes tini to replace launcher_entrypoint as PID 1. I have tested in my environment and it seems to work properly.  This would only be an issue in long running docker jobs that have many steps, eventually you might run out of file descriptors or something.


catto:
related: screwdriver-cd/screwdriver#1403

PR #234, opened 2018-11-07
fix: Update Dockerfile

pranavrc:


Dockerfile
```diff
@@ -75,6 +75,7 @@ RUN set -x \
    && cp /hab/bin/hab /opt/sd/bin/hab \
    # Install Habitat packages
    && /hab/bin/hab pkg install core/bash core/git core/zip core/kmod core/iptables core/docker \
+   && /hab/bin/hab pkg install 7.54.1 core/curl \
```
minz1027: Please add a comment here to point to the artifact bookend to show why we need to pin it here

PR #238, opened 2018-11-15
fix: validate token

s-yoshika:
## Context
Some build pods is running forever although it's build token is invalid and builds is not executed.
```
$ kubectl get pod 
NAME                                READY     STATUS    RESTARTS   AGE
36934-7prwc                         1/1       Running   0          1d
```
The build pod's log is below.
```
[WARN  tini (10)] Tini is not running as PID 1 and isn't registered as a child subreaper.
Zombie processes will not be re-parented to Tini, so zombie reaping won't work.
To fix the problem, use the -s option or set the environment variable TINI_SUBREAPER to register Tini as a child subreaper, or run Tini as PID 1.
2018/11/13 11:01:23 Error getting Build Token 36934: Posting to Build Token: 404 Not Found: Build does not exist
2018/11/13 11:01:23 Starting Build 36934
2018/11/13 11:01:23 No JWT specified. Cannot upload.
Usage of /opt/sd/logservice:
  -api-uri string
    	Base URI for the Screwdriver API ($SD_API_URL)
  -build string
    	ID of the build that is emitting logs ($SD_BUILDID)
  -emitter string
    	Path to the log emitter file (default "/var/run/sd/emitter")
  -lines-per-file int
    	Max number of lines per file when uploading ($SD_LINESPERFILE) (default 1000)
  -store-uri string
    	Base URI for the Screwdriver Store API ($SD_STORE_URL)
  -token string
    	JWT for authenticating with the Store API ($SD_TOKEN)
```
If build token is invalid, the JWT is not issued, so logservice is exit with error log,
but launcher is running without error.

## Objective
This PR add some validation for launcher, and launcher exit safely, like as logservice.

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -595,17 +595,7 @@ func main() {
 
 		if len(token) == 0 {
 			log.Println("Error token is not passed.")
-			os.Exit(0)
-		}
-
-		if len(url) == 0 {
-			log.Println("Error API uri is not specified.")
-			os.Exit(0)
-		}
-
-		if len(storeURL) == 0 {
-			log.Println("Error store uri is not specified.")
-			os.Exit(0)
+			cleanExit()
 		}
 
 		if fetchFlag {
--- a/launch.go
+++ b/launch.go
@@ -593,6 +593,21 @@ func main() {
 			return cli.ShowAppHelp(c)
 		}
 
+		if len(token) == 0 {
+			log.Println("Error token is not passed.")
+			os.Exit(0)
+		}
+
+		if len(url) == 0 {
+			log.Println("Error API uri is not specified.")
+			os.Exit(0)
+		}
+
+		if len(storeURL) == 0 {
+			log.Println("Error store uri is not specified.")
+			os.Exit(0)
+		}
+
 		if fetchFlag {
 			temporalApi, err := screwdriver.New(url, token)
 			if err != nil {
--- a/launch.go
+++ b/launch.go
@@ -594,7 +594,7 @@ func main() {
 		}
 
 		if len(token) == 0 {
-			log.Println("Error token is not passed.")
+			log.Println("Error: token is not passed.")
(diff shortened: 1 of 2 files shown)


launch.go
```diff
@@ -593,6 +593,21 @@ func main() {
 			return cli.ShowAppHelp(c)
 		}
 
+		if len(token) == 0 {
+			log.Println("Error token is not passed.")
+			os.Exit(0)
```
wdstar: We should use `cleanExit()` function.

launch.go
```diff
@@ -593,6 +593,21 @@ func main() {
...
+		}
+
+		if len(url) == 0 {
+			log.Println("Error API uri is not specified.")
+			os.Exit(0)
+		}
+
+		if len(storeURL) == 0 {
+			log.Println("Error store uri is not specified.")
```
wdstar: Don't we have to update build status?

launch.go
s-yoshika: Thanks, I'll fix it.

launch.go
s-yoshika: I fix it to use `exit()`.

s-yoshika: Sorry, It seems that the validations for `url` and `storeUrl` are needless.
These has static default value. So I' remove these validations.
https://github.com/screwdriver-cd/launcher/blob/68a22686724fb00fa40b9024f72dc390765715da/launch.go#L535

s-yoshika:
This PR changes only main function, and I think it can't test main function in golang. Please tell me if there is any good way.

s-yoshika:
I gave up to add this test. I can test to mock cleanExit but then this program goes on without exit.
But in that case the program should exit.

catto:
🙆 I'm okay without adding tests for this change since it has low cost performance.

PR #257, opened 2019-02-14
feat(1483): add default metadata

d2lam:
Related: https://github.com/screwdriver-cd/screwdriver/issues/1483

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -241,11 +241,11 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 	parentBuildIDs := convertToArray(build.ParentBuildID)
 	mergedMeta := map[string]interface{}{
 		"pipelineId": strconv.Itoa(job.PipelineID),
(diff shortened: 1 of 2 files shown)


launch.go
```diff
@@ -239,28 +239,29 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 	metaLog := ""
 
 	parentBuildIDs := convertToArray(build.ParentBuildID)
+	mergedMeta := map[string]interface{}{
```
jithin1987: should we namespace this under build ?

d2lam: not sure, I feel like top level is better since some people might set `meta.build` themselves and it might be confusing

tkyi: wasn't sure what made more sense for preventing people from accidentally setting a special meta field

for meta status and summary I've nested them so people need to set `meta set meta.summary.coverage` or `meta set meta.status.findbugs`

`meta.build` might be better for consistency

PR #258, opened 2019-02-15
fix: create default meta for regular builds

d2lam:
Previously, it will only create default meta for builds that have parentBuild, parentEvent, etc.

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -305,26 +305,26 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 		if event.Meta != nil {
 			mergedMeta = deepMergeJSON(event.Meta, mergedMeta)
 		}
+	}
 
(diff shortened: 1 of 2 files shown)


launch.go
```diff
@@ -248,88 +248,83 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 		"sha":        build.SHA,
 	}
 
-	// If no parent build ID, no parent event, and no event meta, skip fetch meta
-	if len(parentBuildIDs) == 0 && event.ParentEventID == 0 && len(event.Meta) == 0 {
```
d2lam: git diff is weird, the only change is remove this if statement, since now we need to write meta regardless.

jithin1987: Diff settings -> Hide whitespace changes make it readable

d2lam: ohh that's cool! didn't know about this feature! 🎆

PR #270, opened 2019-04-17
feat(1070): evaluate environment variables

d2lam:

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -234,7 +234,6 @@ func Run(path string, env string, emitter screwdriver.Emitter, build screwdriver
 	setupCommands := []string{
 		"set -e",
 		"export PATH=$PATH:/opt/sd",
-		". " + baseEnvFile,
 		// trap EXIT, echo the last step ID and write ENV to /tmp/buildEnv
 		"finish() { " +
 			"EXITCODE=$?; " +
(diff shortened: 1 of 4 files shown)


executor/executor.go
```diff
@@ -233,6 +234,7 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
 	setupCommands := []string{
 		"set -e",
 		"export PATH=$PATH:/opt/sd",
+		". " + baseEnvFile,
```
minz1027: what happened if this cmd failed? e.g. user set some really weird env.

executor/executor.go
```diff
@@ -117,9 +117,9 @@ func doRunCommand(guid, path string, emitter screwdriver.Emitter, f *os.File, fR
 }
 
 // Executes teardown commands
-func doRunTeardownCommand(cmd screwdriver.CommandDef, emitter screwdriver.Emitter, env []string, path, shellBin string, exportFile string) (int, error) {
+func doRunTeardownCommand(cmd screwdriver.CommandDef, emitter screwdriver.Emitter, env string, path, shellBin string, baseEnvFile string, exportFile string) (int, error) {
```
tkyi: can omit `string` for a couple params

```suggestion
func doRunTeardownCommand(cmd screwdriver.CommandDef, emitter screwdriver.Emitter, env, path, shellBin, baseEnvFile, exportFile string) (int, error) {
```

executor/executor.go
d2lam: It will output something like this, just same way it will happen in a shell
```
bash-3.2$ . /tmp/export && echo "bad"
bash: export: `a.b=WJFDS': not a valid identifier
```

d2lam: Moved sourcing to a step so that step is updated

PR #271, opened 2019-04-19
fix(1070): ignore PS1

d2lam:
Currently hanging due to exporting `PS1`. We have this problem with teardown before as well, so need to remove it: https://github.com/screwdriver-cd/launcher/blob/9e5474a626b672a262049eeb9a8dfb2282c9bb06/executor/executor.go#L231

Also move `export PATH` to the createEnvironment part. Otherwise, it will be overwritten. 

Tested in beta and seems working.

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -462,6 +462,8 @@ func createEnvironment(base map[string]string, secrets screwdriver.Secrets, buil
 		}
 	}
 
+	envStrings += "export PATH=$PATH:/opt/sd"
+
 	return envStrings, userShellBin
 }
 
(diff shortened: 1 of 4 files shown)


launch.go
```diff
@@ -462,6 +462,8 @@ func createEnvironment(base map[string]string, secrets screwdriver.Secrets, buil
 		}
 	}
 
+	envStrings += "export PATH=$PATH:/opt/sd"
```
tkyi: does this need to be added first? or would it be okay?

d2lam:
Sorry will open a separate PR

PR #273, opened 2019-04-24
fix: escape newline

d2lam:
Escape newline by wrapping `"`

I couldn't find a better way to do it. Here are a couple ideas but couldn't get it to work:
- Include all env in `exec.Command`, then in `setup-launcher` step, `export -p` to a file. Go through this file and remove OS env. However, this removing needs to be done in between `export -p` and `source`. It's not possible since `osEnv` is stored in `go`. Unless we do some even crazier hack: write osEnv to a separate file then write a script to remove.
- Include buildEnv in `exec.Command`, `export -p` and source this file first. Then append osEnv to  `exec.Command`. But couldn't get this to work since `pty.Start(c)` is done before the `setup-launcher` step.

Please let me know if you have some ideas!

### DIFF
--- a/launch_test.go
+++ b/launch_test.go
@@ -722,32 +722,32 @@ func TestSetEnv(t *testing.T) {
 	defer func() { executorRun = oldExecutorRun }()
 
 	tests := map[string]string{
-		"export PS1":                    "",
-		"export SCREWDRIVER":            "true",
-		"export CI":                     "true",
-		"export CONTINUOUS_INTEGRATION": "true",
-		"export SD_JOB_NAME":            "PR-1",
-		"export SD_PIPELINE_NAME":       "screwdriver-cd/launcher",
-		"export SD_BUILD_ID":            "1234",
-		"export SD_JOB_ID":							"2345",
-		"export SD_EVENT_ID":            "2234",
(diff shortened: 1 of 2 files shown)


launch_test.go
```diff
@@ -892,12 +892,12 @@ func TestCreateEnvironment(t *testing.T) {
 
 	for _, want := range []string{
 		"OSENVWITHEQUALS=foo=bar=",
-		"export SD_TOKEN=1234",
-		"export FOO=bar",
-		"export THINGWITHEQUALS=abc=def",
-		"export secret1=secret1value",
-		"export GETSOVERRIDDEN=override",
-		"export GOPATH=/go/path",
+		"export SD_TOKEN=\"1234\"",
```
jithin1987: these are not same. 

```
C02VD190HV2R:ui jithine$ echo $SD_TOKEN
"1234"
C02VD190HV2R:ui jithine$ export SD_TOKEN=1234
C02VD190HV2R:ui jithine$ echo $SD_TOKEN
1234
```

jithin1987: we can use `eval`

jithin1987:
https://github.com/screwdriver-cd/launcher/pull/274

PR #274, opened 2019-04-25
fix(1590): expandEnv

d2lam:

Doing expansion in `go` rather than doing it in the shell, since we get into different parsing issues. 
Thanks @jithin1987  for suggestions.

### DIFF
--- a/launch_test.go
+++ b/launch_test.go
@@ -251,7 +251,7 @@ func TestMain(m *testing.M) {
 	open = func(f string) (*os.File, error) {
 		return os.Open("data/screwdriver.yaml")
 	}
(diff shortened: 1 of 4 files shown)


launch_test.go
```diff
@@ -873,31 +867,25 @@ func TestCreateEnvironment(t *testing.T) {
...
-		foundEnv[i] = true
-	}
-
-	for _, i := range envArr {
+	for _, i := range env {
 		foundEnv[i] = true
 	}
 
 	for _, want := range []string{
+		"SD_TOKEN=1234",
```
jithin1987: is there an example with expansion ?

d2lam: added

d2lam:
Added a comment about the order here: https://github.com/screwdriver-cd/screwdriver/issues/1070

PR #302, opened 2019-09-12
Try switching to go.mod

scr-oath:
## Context

@jithin1987 pinged me - just trying to switch to modules and see how far I get.

## Objective

glide is out; go mod is in.

## References

https://github.com/golang/go/wiki/Modules#automatic-migration-from-prior-dependency-managers

## License


tkyi:
Probably need these other changes?
https://github.com/screwdriver-cd/meta-cli/pull/18/files#diff-e917068babdeb1068c2aa1a34cb665ca

scr-oath:
Hmm… what version of go do you have? It doesn't have mod?!?

tkyi:
Probably need to use this hack as well: https://github.com/russross/blackfriday/issues/500

scr-oath:
Anyone know how to fix the `test-release` step? The rest of it seems to build, but something about dynamic libraries for mac having issues.

tkyi:
Could you try adding `--debug` to https://github.com/screwdriver-cd/launcher/blob/master/screwdriver.yaml#L25?

scr-oath:
Added --debug

.goreleaser.yml
```diff
@@ -8,7 +8,7 @@ builds:
       - amd64
     # Include the default settings from https://goreleaser.com/#builds
     # Also include static compilation
-    ldflags: -d -s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}} -extldflags "-static"
+    ldflags: -s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}} -extldflags "-static"
```
dmitris: reference from https://golang.org/cmd/link/:

> -d
	Disable generation of dynamic executables.
	The emitted code is the same in either case; the option
	controls only whether a dynamic header is included.
	The dynamic header is on by default, even without any
	references to dynamic libraries, because many common
	system tools now assume the presence of the header.

go.mod
```diff
@@ -0,0 +1,19 @@
...
+
+go 1.12
+
+require (
+	github.com/mattn/go-colorable v0.1.2 // indirect
+	github.com/mattn/go-isatty v0.0.9 // indirect
+	github.com/myesui/uuid v1.0.0 // indirect
+	github.com/peterbourgon/mergemap v0.0.0-20130613134717-e21c03b7a721
+	gopkg.in/fatih/color.v1 v1.7.0
+	gopkg.in/kr/pty.v1 v1.1.8
```
dmitris: the https://github.com/kr/pty repo says: "New clients should use the new module import path github.com/creack/pty", should we migrate to it?

go.mod
```diff
@@ -0,0 +1,19 @@
...
+	gopkg.in/myesui/uuid.v1 v1.0.0
+	gopkg.in/stretchr/testify.v1 v1.4.0 // indirect
+	gopkg.in/urfave/cli.v1 v1.20.0
+)
+
+replace gopkg.in/kr/pty.v1 => github.com/kr/pty v1.1.8
```
dmitris: are these `replace` statements required?  If so, I would group them into one multiline`replace` block

scr-oath: I did that, but go mod tidy puts them back - I've been under the impression that, with golang, one should always be able to say go fmt ./... or go mod tidy and that is the de facto style guide.

.goreleaser.yml
scr-oath: So it's ok then… Wonder why it worked with go 1.11 but not newer ones…

go.mod
scr-oath: Done

.goreleaser.yml
```diff
@@ -8,7 +8,7 @@ builds:
       - amd64
     # Include the default settings from https://goreleaser.com/#builds
     # Also include static compilation
-    ldflags: -s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}} -extldflags "-static"
+    ldflags: -d -s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}} -extldflags "-static"
```
jithin1987: lets remove `ldflags`. `-d & extldflags` is what we add extra, which I don't this is needed.

PR #320, opened 2019-12-17
feat(1314): Add Local mode

kkisic:
## Context
Add local-mode for `sd-local`

## Objective
- Add some options for local-mode
  - `--local-mode`
  - `--local-build-json`
  - `--local-job-name`
- Add `local_run.sh`
- Add localApi for local-mode

## References
https://github.com/screwdriver-cd/screwdriver/issues/1314

## License

### DIFF
--- a/Docker/local_run.sh
+++ b/Docker/local_run.sh
@@ -1,5 +1,6 @@
 #!/bin/sh
 
+#Because launcher doesn't use buildID in local-mode, pass 0 to launcher as dummy.
 /opt/sd/launch --local-mode --local-build-json "$1" --local-job-name "$2" --api-uri "$3" --emitter /sd/emitter 0 &
 /opt/sd/logservice --local-mode --build-log-file "$4" --emitter /sd/emitter &
 wait $(jobs -p)
--- a/Docker/local_run.sh
+++ b/Docker/local_run.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+/opt/sd/launch --local-mode --local-build-json "$1" --local-job-name "$2" --api-uri "$3" --emitter /sd/emitter 0 &
+/opt/sd/logservice --local-mode --build-log-file "$4" --emitter /sd/emitter &
+wait $(jobs -p)
--- a/Docker/local_run.sh
+++ b/Docker/local_run.sh
@@ -1,6 +1,10 @@
 #!/bin/sh
 
-#Because launcher doesn't use buildID in local-mode, pass 0 to launcher as dummy.
-/opt/sd/launch --local-mode --local-build-json "$1" --local-job-name "$2" --api-uri "$3" --emitter /sd/emitter 0 &
-/opt/sd/logservice --local-mode --build-log-file "$4" --emitter /sd/emitter &
+# Create FIFO for emitter
+mkdir -p /sd
+mkfifo -m 666 /sd/emitter
+
+# Because launcher doesn't use buildID in local-mode, pass 0 to launcher as dummy.
+/opt/sd/launch --local-mode --local-build-json "$1" --local-job-name "$2" --api-uri "$3" --store-uri "$4" --emitter /sd/emitter 0 &
+/opt/sd/logservice --local-mode --build-log-file "$5" --emitter /sd/emitter &
 wait $(jobs -p)
--- a/launch.go
+++ b/launch.go
@@ -698,7 +698,7 @@ func main() {
 
 		log.Printf("cache strategy, directories (pipeline, job, event), compress, md5check, maxsize: %v, %v, %v, %v, %v, %v, %v \n", cacheStrategy, pipelineCacheDir, jobCacheDir, eventCacheDir, cacheCompress, cacheMd5Check, cacheMaxSizeInMB)
 
-		if len(token) == 0 {
+		if !isLocal && len(token) == 0 {
 			log.Println("Error: token is not passed.")
 			cleanExit()
 		}
--- a/launch.go
+++ b/launch.go
@@ -50,7 +50,7 @@ const DefaultTimeout = 90 // 90 minutes
 
 // exit sets the build status and exits successfully
 func exit(status screwdriver.BuildStatus, buildID int, api screwdriver.API, metaSpace string) {
-	if api != nil {
+	if api != nil && !api.IsLocal() {
 		var metaInterface map[string]interface{}
 
 		log.Printf("Loading meta from %q/meta.json", metaSpace)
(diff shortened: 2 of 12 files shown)


launch.go
```diff
@@ -653,6 +665,18 @@ func main() {
...
 			Usage: "Cache allowed max size in mb",
 		},
+		cli.BoolFlag{
+			Name:  "local-mode",
+			Usage: "Enable local mode",
+		},
+		cli.StringFlag{
+			Name:  "local-build-json",
+			Usage: "Build information for local mode",
+		},
```
kkisic: Pass JSON in the same format as payload from `/v4/builds/{id}`

Docker/local_run.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+/opt/sd/launch --local-mode --local-build-json "$1" --local-job-name "$2" --api-uri "$3" --emitter /sd/emitter 0 &
```
MysticDoll: What is this `0` argument?

kkisic: Because launcher doesn't use `buildID` in local-mode, pass `0` to launcher as dummy.

tk3fftk: I think we should add a comment for this.

launch.go
tk3fftk: Do we need to copy this comment to a code comment?

launch.go
```diff
@@ -280,7 +288,7 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 
 	// Create meta space
 	log.Printf("Creating Meta Space in %v", metaSpace)
```
tk3fftk: This log will be always printed even if we use local mode and meta space is not created.

launch.go
kkisic: Add comment

Docker/local_run.sh
kkisic: Added comment

launch.go
kkisic: I moved this log into `createMetaSpace` like any other logs about `Meta`

launch.go
```diff
@@ -147,15 +147,24 @@ func createWorkspace(rootDir string, srcPaths ...string) (Workspace, error) {
 	return w, nil
 }
 
-func createMetaSpace(metaSpace string) error {
+func createMetaSpace(metaSpace string, isLocal bool) error {
```
tk3fftk: I think if we can pass `meta` from `sd-local`, we should create a `metafile` to use the passed `meta` in a local build.
it's referenced this line: https://github.com/screwdriver-cd/screwdriver/pull/1884/files#diff-49f99a06253f0a9a292e89f6a469b9e4R46

kkisic: I forgot that `sd-local` support `SD_META_DIR` and `SD_META_PATH`.
I fixed it.

screwdriver/screwdriver_local.go
```diff
@@ -0,0 +1,100 @@
...
+func (a localApi) BuildFromID(buildID int) (build Build, err error) {
+	return a.localBuild, nil
+}
+
+func (a localApi) EventFromID(eventID int) (event Event, err error) {
+	event = Event{
+		0,
+		make(map[string]interface{}),
+		0,
+	}
```
yuichi10: It is hard to understand what 0 is.
I think it is better to add field name.

screwdriver/screwdriver_local.go
```diff
@@ -0,0 +1,100 @@
...
+	return event, nil
+}
+
+func (a localApi) JobFromID(jobID int) (job Job, err error) {
+	job = Job{
+		0,
+		0,
+		a.jobName,
+		0,
+	}
```
yuichi10: Same here. I think it is better to add field name.

screwdriver/screwdriver_local.go
```diff
@@ -0,0 +1,100 @@
...
+
+	return job, nil
```
yuichi10: [nits]
maybe you do not need write `job, nil`.
Just return is enough

screwdriver/screwdriver_local.go
kkisic: I fixed it.

screwdriver/screwdriver_local.go
kkisic: I fixed it.

launch.go
```diff
@@ -403,12 +403,14 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 
 	apiURL, _ := api.GetAPIURL()
 
+	isCI := strconv.FormatBool(!api.IsLocal())
```
yuichi10: IsLocal() method is used only here?
If so, is it better pass isLocal as an argument?
Then the API interface does not need to have IsLocal function.

kkisic: I didn't want to add more arguments because there were too many arguments for `launch`

screwdriver/screwdriver_local.go
kkisic: Although it is written redundantly in `screwdriver.go`, should I omit it?

launch.go
yuichi10: I feel strange the API have IsLocal function. 
I think the good point to use Interface is you do not need to specify the interface is.

yuichi10: If the arguments are too many. Maybe we can use struct for the arguments.

kkisic: Remove `API.IsLocal()` and add argument to `launch` and `launchAction`

screwdriver/screwdriver_local.go
yuichi10: I see.
I think we do not need to fix this.

launch_test.go
```diff
@@ -111,6 +111,9 @@ func mockAPI(t *testing.T, testBuildID, testJobID, testPipelineID int, testStatu
 			}
 			return "foobar", nil
 		},
+		isLocal: func() bool {
```
yuichi10: I think we can delete this.

launch_test.go
```diff
@@ -126,6 +129,7 @@ type MockAPI struct {
 	getAPIURL         func() (string, error)
 	getCoverageInfo   func() (screwdriver.Coverage, error)
 	getBuildToken     func(buildID int, buildTimeoutMinutes int) (string, error)
+	isLocal           func() bool
```
yuichi10: We can delete this.

launch_test.go
kkisic: I fixed it.

launch_test.go
kkisic: I fixed it.

PR #325, opened 2020-01-24
fix: change cacheMaxSizeInMB as StringFlag and parse as int64

adong:
## Context

Currently, if our `./Docker/run.sh` pass empty string `""` to `cacheMaxSizeInMB`, will cause `parse error`

image

## Objective

This PR will relax and do a string to `int64` conversion for `cacheMaxSizeInMB` 

## License

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -651,7 +651,7 @@ func main() {
 			Name:  "cache-md5check",
 			Usage: "Do md5 check",
 		},
-		cli.Int64Flag{
+		cli.StringFlag{
 			Name:  "cache-max-size-mb",
 			Usage: "Cache allowed max size in mb",
 		},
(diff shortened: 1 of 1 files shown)


launch.go
```diff
@@ -688,7 +688,7 @@ func main() {
 		eventCacheDir := c.String("event-cache-dir")
 		cacheCompress, _ := strconv.ParseBool(c.String("cache-compress"))
 		cacheMd5Check, _ := strconv.ParseBool(c.String("cache-md5check"))
-		cacheMaxSizeInMB := c.Int64("cache-max-size-mb")
+		cacheMaxSizeInMB, _ := strconv.ParseInt(c.String("cache-max-size-mb"), 10, 64)
```
parthasl: conversion is not needed, by default it will be 0

jithine:
No longer needed with https://github.com/screwdriver-cd/launcher/pull/326/files

PR #326, opened 2020-01-24
fix(1928): store-cli - runtime error 10000 thread limit

parthasl:
## Context

If files to be cached are more than 15000, store-cli throws runtime error 10000-thread limit and the cache-max-size-mb argument to be read as a string instead of Int64 

## Objective

This PR 1. gets the latest store-cli which has changes related to reduce md5check concurrency process limit to 3000 from 10000 and 2. update cache-max-size-mb argument to be read as a string instead of Int64 to handle docker error when sd is run locally.

## References

[1928](https://github.com/screwdriver-cd/screwdriver/issues/1928)

## License

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -654,7 +654,7 @@ func main() {
 		cli.StringFlag{
 			Name:  "cache-max-size-mb",
 			Usage: "Cache allowed max size in mb",
-			Value:  "0",
+			Value: "0",
 		},
 		cli.BoolFlag{
 			Name:  "local-mode",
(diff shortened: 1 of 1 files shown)


parthasl:
> 1000 change seems to be gone ?
> Also does this mean a typecast needs to be done like this https://github.com/screwdriver-cd/launcher/pull/325/files#diff-338987301073a98ce4db3faf2cc35204R691

typecast is not needed. c.Int64 accepts string and returns Int64 value

PR #328, opened 2020-01-27
fix(1926): Add SD_UI_BUILD_URL

anilkumarmyla:
## Context

Fixes https://github.com/screwdriver-cd/screwdriver/issues/1926

## License

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -432,7 +432,7 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 		"SD_BUILD_URL":           apiURL + "builds/" + strconv.Itoa(buildID),
 		"SD_STORE_URL":           fmt.Sprintf("%s/%s/", storeURL, "v1"),
 		"SD_UI_URL":              fmt.Sprintf("%s/", uiURL),
-		"SD_UI_BUILD_URL":        uiURL + "/builds/" + strconv.Itoa(buildID),
+		"SD_UI_BUILD_URL":        fmt.Sprintf("%s/pipelines/%s/builds/%s", uiURL, strconv.Itoa(job.PipelineID), strconv.Itoa(buildID)),
 		"SD_TOKEN":               buildToken,
 		"SD_CACHE_STRATEGY":      cacheStrategy,
 		"SD_PIPELINE_CACHE_DIR":  pipelineCacheDir,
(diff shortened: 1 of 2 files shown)


launch.go
```diff
@@ -432,6 +432,7 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 		"SD_BUILD_URL":           apiURL + "builds/" + strconv.Itoa(buildID),
 		"SD_STORE_URL":           fmt.Sprintf("%s/%s/", storeURL, "v1"),
 		"SD_UI_URL":              fmt.Sprintf("%s/", uiURL),
+		"SD_UI_BUILD_URL":        uiURL + "/builds/" + strconv.Itoa(buildID),
```
tkyi:
```suggestion
		"SD_UI_BUILD_URL":        uiURL + "builds/" + strconv.Itoa(buildID),
```

anilkumarmyla: @tkyi my original commit was what you suggested, it failed the unit test since if you check the line above `SD_UI_URL` is initialized by suffixing a `/` at end of uiURL

launch_test.go
```diff
@@ -745,6 +745,7 @@ func TestSetEnv(t *testing.T) {
 		"SD_BUILD_URL":           "https://api.screwdriver.cd/v4/builds/1234",
 		"SD_STORE_URL":           "https://store.screwdriver.cd/v1/",
 		"SD_UI_URL":              "https://screwdriver.cd/",
+		"SD_UI_BUILD_URL":        "https://screwdriver.cd/builds/1234",
```
jithine: If the intention is to use this end point to further construct a link to Artifacts url, tgeb this won't give you want you want if. You need the full url eg: `https://screwdriver.cd/pipelines/300461/builds/6018884` You do have the pipeline id, so use that to construct full URL.

launch.go
jithine: as per line 434 SD_UI_URL already has a trailing slash 🤔

launch_test.go
anilkumarmyla: there is some magic happening where opening url
https://screwdriver.cd/builds/6018884 would take me directly to the page even without the pipelineId. 

Is /pipelines/{pipelineId} the standard url, i didn't find any `pipelines` string in launch.go 🤔

jithine: it does, but it doesn't support more path params. so with this `$SD_UI_BUILD_URL/artifacts/result.html` won't work.

jithine: you have https://github.com/screwdriver-cd/launcher/pull/328/files#diff-338987301073a98ce4db3faf2cc35204R419, so `pipeline id` is available

anilkumarmyla: `/pipelines/{pipelineId}` is something that might be related to the UI component, is it ok to generate this url here?

anilkumarmyla: @jithine made the changes per your feedback to include `/pipelines/{pipelineId}`

PR #331, opened 2020-02-10
fix(1314): Fix about local-mode

kkisic:
## Context
`sd-local` mounts workspace from host filesystem, but launcher check it has already existed, and if so launcher throws error.
To avoid this, this PR makes launcher not to throw that error.
And, we forgot to copy `local_run.sh` to Docker Image.
And furthermore, we want to use `/sd/workspace/src/{dummyScmHost}/{dummyScmOrg}/{dummyScmRepo}` as source directory in local-mode

## Objective
- make `createWorkspace` function not to throw error about `path already exists` in local-mode
- copy `local_run.sh` to Docker Image
- fix dummy `Pipeline` information

## References
screwdriver-cd/screwdriver#1314

## License

### DIFF
--- a/launch_test.go
+++ b/launch_test.go
@@ -382,7 +382,7 @@ func TestCreateWorkspace(t *testing.T) {
 		madeDirs[path] = perm
 		return nil
 	}
-	workspace, err := createWorkspace(TestWorkspace, "screwdriver-cd", "launcher")
+	workspace, err := createWorkspace(false, TestWorkspace, "screwdriver-cd", "launcher")
 
 	if err != nil {
 		t.Errorf("Unexpected error creating workspace: %v", err)
@@ -454,7 +454,7 @@ func TestCreateWorkspaceBadStat(t *testing.T) {
 
 	wantWorkspace := Workspace{}
 
-	workspace, err := createWorkspace(TestWorkspace, "screwdriver-cd", "launcher")
+	workspace, err := createWorkspace(false, TestWorkspace, "screwdriver-cd", "launcher")
 
 	if err.Error() != "Cannot create workspace path \"/sd/workspace/src/screwdriver-cd/launcher\", path already exists." {
 		t.Errorf("Error is wrong, got %v", err)
--- a/launch_test.go
+++ b/launch_test.go
@@ -465,6 +465,31 @@ func TestCreateWorkspaceBadStat(t *testing.T) {
 	}
 }
 
+func TestCreateWorkspaceBadStatLocal(t *testing.T) {
+	oldStat := stat
+	defer func() { stat = oldStat }()
+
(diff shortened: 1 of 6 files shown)


kkisic:
There is still something we need to fix, so do not merge this PR.

screwdriver/screwdriver_local.go
```diff
@@ -54,8 +54,8 @@ func (a localApi) JobFromID(jobID int) (job Job, err error) {
 func (a localApi) PipelineFromID(pipelineID int) (pipeline Pipeline, err error) {
 	pipeline = Pipeline{
 		ID:      0,
-		ScmRepo: ScmRepo{"screwdriver-cd/screwdriver"},
-		ScmURI:  "github.com:123456:master",
```
sugarnaoming: Fixed to use `/sd/workspace/src/screwdriver.cd/local-build/src` as `sourceDir` in `sd-local`.

screwdriver/screwdriver_local.go
```diff
@@ -54,8 +54,8 @@ func (a localApi) JobFromID(jobID int) (job Job, err error) {
...
+		ScmRepo: ScmRepo{"sd-local/local-build"},
+		ScmURI:  "screwdriver.cd:123456:master",
```
sugarnaoming: Fixed to use `/sd/workspace/src/screwdriver.cd/sd-local/local-build` as `sourceDir` in `sd-local`.

screwdriver/screwdriver_local.go
```diff
@@ -54,8 +54,8 @@ func (a localApi) JobFromID(jobID int) (job Job, err error) {
 func (a localApi) PipelineFromID(pipelineID int) (pipeline Pipeline, err error) {
 	pipeline = Pipeline{
 		ID:      0,
-		ScmRepo: ScmRepo{"screwdriver-cd/screwdriver"},
-		ScmURI:  "github.com:123456:master",
+		ScmRepo: ScmRepo{"local-build/src"},
```
tk3fftk: This may be `sd-local/local-build`.

launch_test.go
```diff
@@ -454,7 +454,7 @@ func TestCreateWorkspaceBadStat(t *testing.T) {
 
 	wantWorkspace := Workspace{}
 
-	workspace, err := createWorkspace(TestWorkspace, "screwdriver-cd", "launcher")
+	workspace, err := createWorkspace(false, TestWorkspace, "screwdriver-cd", "launcher")
```
yuichi10: Don't we need the test case when  isLocal is true.

kkisic: Thx.
I added test for that case.

kkisic:
> There is still something we need to fix, so do not merge this PR.

Would you please review, because all modifications has done

PR #335, opened 2020-03-25
feat(818): push image metrics to prometheus

parthasl:
## Context

Top images to be cached in the node are not available as part of the node warm-up.

## Objective

Send image metrics to Prometheus via pushgateway configuration. These metrics will be used to query top n images and cached in node as part of the node warm-up. Environment variables are passed from executor-k8s.

When Docker/run.sh is called via executor-k8s-vm, it will skip metrics push to Prometheus. It's already implemented via hyperctl-docker scripts.

## References

[Evaluate Kata Containers over HyperD](https://github.com/screwdriver-cd/screwdriver/issues/818)
[Executor-k8s PR-117](https://github.com/screwdriver-cd/executor-k8s/pull/117)

## License

### DIFF
--- a/Docker/run.sh
+++ b/Docker/run.sh
@@ -2,7 +2,7 @@
 
 # push data to prometheus via pushgateway url
 function pushToPrometheus {
-  curl -m 10 --data-binary @- "$PUSHGATEWAY_URL/metrics/job/containerd$1" || true
+  curl -s -m 10 --data-binary @- "$PUSHGATEWAY_URL/metrics/job/containerd$1" &>/dev/null &
 }
 
 # Get push gateway url and container image from env variable
--- a/Docker/run.sh
+++ b/Docker/run.sh
@@ -6,7 +6,7 @@ function pushToPrometheus {
 }
 
 # Get push gateway url and container image from env variable
-if [[ ! -z "$PUSHGATEWAY_URL" ]] && [[ ! -z "$CONTAINER_IMAGE" ]] && [[ !z "$SD_PIPELINE_ID" ]]; then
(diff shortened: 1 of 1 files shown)


Docker/run.sh
```diff
@@ -1,4 +1,17 @@
 #!/bin/sh
 
+# push data to prometheus via pushgateway url
+function pushToPrometheus {
+  curl -m 10 --data-binary @- "$PUSHGATEWAY_URL/metrics/job/containerd$1" || true
```
jithine: can `pushToPrometheus` be called as background process  ?

Docker/run.sh
```diff
@@ -1,4 +1,17 @@
 #!/bin/sh
 
+# push data to prometheus via pushgateway url
+function pushToPrometheus {
+  curl -s -m 10 --data-binary @- "$PUSHGATEWAY_URL/metrics/job/containerd$1" &>/dev/null &
+}
+
+# Get push gateway url and container image from env variable
+if [[ ! -z "$PUSHGATEWAY_URL" ]] && [[ ! -z "$CONTAINER_IMAGE" ]] && [[ !z "$SD_PIPELINE_ID" ]]; then
```
tk3fftk: I think `[[ ]]` is a bash (or modern shells) feature.

https://serverfault.com/questions/52034/what-is-the-difference-between-double-and-single-square-brackets-in-bash

parthasl: yes tested its working.

tk3fftk: I see, thanks. 
And I found typo: `[[ !z "$SD_PIPELINE_ID" ]]` -> `[[ ! -z "$SD_PIPELINE_ID" ]]` ?

parthasl: uwc, fixed typo

PR #355, opened 2020-06-02
fix: expose step exit code and pass to teardown

DekusDenial:
## Context

It will be useful for teardown steps to be aware of the build status thus far inferred from the exit codes from previously executed steps.

## Objective

To include previous steps exit code as env var inside a teardown step

## References

https://github.com/screwdriver-cd/screwdriver/issues/2103

## License

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -151,9 +151,9 @@ func doRunCommand(guid, path string, emitter screwdriver.Emitter, f *os.File, fR
 }
 
 // Executes teardown commands
-func doRunTeardownCommand(cmd screwdriver.CommandDef, emitter screwdriver.Emitter, path, shellBin, exportFile, sourceDir string) (int, error) {
+func doRunTeardownCommand(cmd screwdriver.CommandDef, emitter screwdriver.Emitter, path, shellBin, exportFile, sourceDir string, stepExitCode int) (int, error) {
 	shargs := []string{"-e", "-c"}
-	cmdStr := "export PATH=$PATH:/opt/sd && " +
+	cmdStr := "export PATH=$PATH:/opt/sd SD_STEP_EXIT_CODE=" + strconv.Itoa(stepExitCode) + " && " +
 		"START=$(date +'%s'); while ! [ -f " + exportFile + " ] && [ $(($(date +'%s')-$START)) -lt " + strconv.Itoa(WaitTimeout) + " ]; do sleep 1; done; " +
 		"if [ -f " + exportFile + " ]; then set +e; . " + exportFile + "; set -e; fi; " +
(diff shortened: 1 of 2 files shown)


executor/executor.go
```diff
@@ -151,9 +151,9 @@ func doRunCommand(guid, path string, emitter screwdriver.Emitter, f *os.File, fR
 }
 
 // Executes teardown commands
-func doRunTeardownCommand(cmd screwdriver.CommandDef, emitter screwdriver.Emitter, path, shellBin, exportFile, sourceDir string) (int, error) {
+func doRunTeardownCommand(cmd screwdriver.CommandDef, emitter screwdriver.Emitter, path, shellBin, exportFile, sourceDir string, aggExitCode int) (int, error) {
 	shargs := []string{"-e", "-c"}
-	cmdStr := "export PATH=$PATH:/opt/sd && " +
+	cmdStr := "export PATH=$PATH:/opt/sd AGG_EXIT_CODE=" + strconv.Itoa(aggExitCode) + " && " +
```
jithine, line 156: can we call this `SD_STEP_EXIT_CODE` and document that any non-zero value indicates build failure. 

can this be translated to build status ?

executor/executor.go
```diff
@@ -358,7 +361,8 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
 			return fmt.Errorf("Updating step start %q: %v", cmd.Name, err)
 		}
 
-		code, cmdErr = doRunTeardownCommand(cmd, emitter, path, shellBin, exportFile, sourceDir)
+		code, cmdErr = doRunTeardownCommand(cmd, emitter, path, shellBin, exportFile, sourceDir, aggExitCode)
+		aggExitCode = code | aggExitCode
```
jithine, line 365: why aggregate ? can't we just expose the latest non-zero step exit code ?

executor/executor.go
DekusDenial: yes, i can rename it.
no, should leave it up to each teardown to interpret it differently, and teardown step itself can fail

executor/executor.go
DekusDenial: sure

parthasl:
looks good. we expect teardown steps to exit smoothly, for any errors as well correct?

DekusDenial:
> looks good. we expect teardown steps to exit smoothly, for any errors as well correct? and we want to capture only last step exit code and not other details? are we sending this as metric to Yamas?

My use case is for last sd-teardown step to know what presumably going to be the build status, but this also let teardown steps to decide what to do based on failing status from previous user steps. And yes, we are sending certain event based on the outcome too.

PR #369, opened 2020-09-25
feat: Run teardown steps on abort build

pritamstyz4ever:
## Context

Currently if a build is aborted the teardown steps don't run and hence generated artifacts are not available in this case.

## Objective

This PR adds change to capture terminate signal and run the teardown step in the current build

## References

https://github.com/screwdriver-cd/screwdriver/issues/1125

## License

### DIFF
--- a/Docker/launcher_entrypoint.sh
+++ b/Docker/launcher_entrypoint.sh
@@ -9,7 +9,7 @@ clean_up () {
 }
 
 # Trap these SIGNALs and run teardown
-trap 'clean_up $@'  SIGINT SIGTERM EXIT SIGQUIT SIGHUP
+trap 'clean_up $@' HUP INT QUIT TERM EXIT
 
 I_AM_ROOT=false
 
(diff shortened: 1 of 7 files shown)


Docker/run.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
parthasl, line 1:
```suggestion
#!/bin/bash -e
```

Docker/launcher_entrypoint.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
parthasl, line 1:
```suggestion
#!/bin/sh
```

parthasl:
@pritamstyz4ever better to move this SIGTERM logic inside executor.Run. Let's revisit this in the next iteration.

parthasl (approved):

PR #370, opened 2020-10-02
fix: exit launcher script on error

jithine:
## Context

Launcher entry point is not exiting on error, leading to Kubernetes build pods stuck forever.

## Objective

Exit launcher entry script on error

## License


parthasl (approved):
👍

PR #371, opened 2020-10-15
feat(2084): Add env for periodic build

kumada626:
## Context

Add `SD_SCHEDULING_BUILD` for periodic build.
If the build is triggered by scheduler, `SD_SCHEDULING_BUILD` is `true`.
Otherwise, `SD_SCHEDULING_BUILD` is `false`.
If there are any other appropriate environment name, I'll fix the environment name.

## Objective

If an event is created by scheduler, event's creator username is `sd:scheduler`.
https://github.com/screwdriver-cd/queue-service/blob/415aa19006f30be7e69753c8736eb03bf3d9c1ea/plugins/queue/scheduler.js#L46-L54
In such case, set env `SD_SCHEDULING_BUILD` true.

## References

https://github.com/screwdriver-cd/screwdriver/issues/2084

## License

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -538,7 +538,7 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 		"SD_CACHE_MD5CHECK":       fmt.Sprintf("%v", cacheMd5Check),
 		"SD_CACHE_MAX_SIZE_MB":    fmt.Sprintf("%v", cacheMaxSizeInMB),
 		"SD_CACHE_MAX_GO_THREADS": fmt.Sprintf("%v", cacheMaxGoThreads),
-		"SD_SCHEDULING_BUILD":     isScheduler,
+		"SD_SCHEDULED_BUILD":     isScheduler,
 	}
 
 	// Add coverage env vars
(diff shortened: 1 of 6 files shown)


wahapo (approved):

ibu1224 (approved):

screwdriver/screwdriver.go
```diff
@@ -186,6 +186,7 @@ type Event struct {
 	ID            int                    `json:"id"`
 	Meta          map[string]interface{} `json:"meta"`
 	ParentEventID int                    `json:"parentEventId"`
+	Creator       map[string]interface{} `json:"creator"`
```
cappyzawa, line 189: Isn't the only type of Creator a string?

```suggestion
	Creator       map[string]string `json:"creator"`
```

kumada626: You are right. I fixed type from interface to string.

cappyzawa (approved):

launch.go
```diff
@@ -536,6 +538,7 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 		"SD_CACHE_MD5CHECK":       fmt.Sprintf("%v", cacheMd5Check),
 		"SD_CACHE_MAX_SIZE_MB":    fmt.Sprintf("%v", cacheMaxSizeInMB),
 		"SD_CACHE_MAX_GO_THREADS": fmt.Sprintf("%v", cacheMaxGoThreads),
+		"SD_SCHEDULING_BUILD":     isScheduler,
```
tkyi, line 541: maybe `SD_SCHEDULED_BUILD` or 
not sure if it would be better if we could have something like `SD_TRIGGER` or `SD_START_FROM` and it would be `scheduler` or `commit` or `manual`

kumada626: Fix env name to `SD_SCHEDULED_BUILD`.

tkyi (approved):

PR #378, opened 2020-12-01
fix(2278): Update external files for join job meta

tkyi:
## Context

Users with external join jobs are finding that rerunning an old commit is pulling in stale external metadata.

## Objective

This PR will check if parent build is from an external pipeline or not. If so, write meta to external file as well as merge meta to main build's meta.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2278

## License

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -270,56 +270,90 @@ func createMetaSpace(metaSpace string) error {
 }
 
 func writeMetafile(metaSpace, metaFile, metaLog string, mergedMeta map[string]interface{}) error {
-	metaByte := []byte("")
-	log.Println("Marshalling Merged Meta JSON")
-	metaByte, err := marshal(mergedMeta)
+	var metaPath = metaSpace + "/" + metaFile
 
-	if err != nil {
-		return fmt.Errorf("Parsing Meta JSON: %v", err)
+	_, err := os.Stat(metaPath)
+
+	// If file doesn't already exist, write to new file
+	if os.IsNotExist(err) {
+		metaByte := []byte("")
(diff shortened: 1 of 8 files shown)


launch.go
```diff
@@ -270,19 +270,90 @@ func createMetaSpace(metaSpace string) error {
...
+
+		if err != nil {
+			return fmt.Errorf("Parsing Meta JSON: %v", err)
+		}
+
+		err = writeFile(metaPath, metaByte, 0666)
+		if err != nil {
+			return fmt.Errorf("Writing Parent %v Meta JSON: %v", metaLog, err)
+		}
+	} else {
```
jithine, line 291: will this ever happen ? It's a new build, and no external fetch has happened yet ?

launch.go
```diff
@@ -270,19 +270,90 @@ func createMetaSpace(metaSpace string) error {
...
+		err = writeFile(metaPath, metaByte, 0666)
+		if err != nil {
+			return fmt.Errorf("Writing Parent %v Meta JSON: %v", metaLog, err)
+		}
+	}
+
+	return nil
+}
 
+// SetExternalMeta checks if parent build is external and sets meta in external file accordingly
```
jithine, line 326: is writing to external meta file enough? meta cli looks at `meta key` first https://github.com/screwdriver-cd/meta-cli/blob/9cfbdc987b7c55e6a03559f0d793d1f321124fa4/meta.go#L77

So even if we update the external file, won't `meta key` still have old value ? I think updating the `meta key` is also needed

tkyi:
closing in favor of https://github.com/screwdriver-cd/launcher/pull/382

PR #383, opened 2020-12-15
fix(2278): Reset local version of external meta

tkyi:
## Context

We need to reset the local version of external meta so it can be fetched from the external file, since meta-cli will check local version first, then check external file: https://github.com/screwdriver-cd/meta-cli/blob/master/meta.go#L86

## Objective

This PR checks if a local version of the desired external meta exists, then sets it to `null`. 

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2278

## License

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -308,9 +308,17 @@ func SetExternalMeta(api screwdriver.API, pipelineID, parentBuildID int, mergedM
 			if join {
 				resultMeta = deepMergeJSON(parentBuild.Meta, resultMeta)
 			}
-			externalMetaKey := "sd." + strconv.Itoa(parentJob.PipelineID) + "." + parentJob.Name
-			// delete stored external meta key
-			delete(resultMeta, externalMetaKey)
+
+			// delete local version of external meta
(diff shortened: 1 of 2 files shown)


launch.go
```diff
@@ -308,9 +308,17 @@ func SetExternalMeta(api screwdriver.API, pipelineID, parentBuildID int, mergedM
...
-			// delete stored external meta key
-			delete(resultMeta, externalMetaKey)
+
+			// set stored external meta key to "null"
+			pIDString := strconv.Itoa(parentJob.PipelineID)
+			pjn := parentJob.Name
+			if sdMeta, ok := resultMeta["sd"]; ok {
+				if externalPipelineMeta, ok := sdMeta.(map[string]interface{})[pIDString]; ok {
+					if _, ok := externalPipelineMeta.(map[string]interface{})[pjn]; ok {
+						externalPipelineMeta.(map[string]interface{})[pjn] = "null"
```
jithine, line 318: I think this can be `delete(externalPipelineMeta, pjn)` to remove the job key from `externalPipelineMeta`

jithine (approved):

PR #388, opened 2021-01-12
feat: add export env vars

pritamstyz4ever:
## Context

Env variables not available in teardown steps

## Objective

This PR adds the export to env vars in teardown step

## License


executor/executor.go
```diff
@@ -393,6 +394,28 @@ func RunTeardownOnAbort(path string, env []string, emitter screwdriver.Emitter,
...
+		"export PATH=$PATH:/opt/sd",
+		// trap EXIT, echo the last step ID and write ENV to /tmp/buildEnv
+		"finish() { " +
+			"EXITCODE=$?; " +
+			exportEnvCmd +
+			"echo $SD_STEP_ID $EXITCODE; }", //mv newfile to file
+		"trap finish EXIT;\necho ;\n",
+	}
+
+	setupReader := bufio.NewReader(f)
```
jithine, line 414: will this work when abort happens ?

by the time this function runs https://github.com/screwdriver-cd/launcher/blob/cd4a62a4e0e156fc15576ac21989ce5eb01ab12b/launch.go#L667 , is the current running step already aborted ?

pritamstyz4ever: yes, teardown is triggered via the trap handler

parthasl: what happens if they create/update the environment variable and want to use the updated value in the teardown step? it will not be possible and might need to revisit later.

jithine: Looking at this build https://beta.cd.screwdriver.cd/pipelines/6512/builds/118185/steps/teardown-step1 
This PR still hasn't fixed the issue, right ? `ENV_VAR_TEST` should have been available ? 

I think need to have a different logic for env export which complements `trap`. 
At end of each step run, export the env data to a separate file and use this file to restore `env`

pritamstyz4ever: seems this is an existing behavior env var exported in one of the steps is not available in the other.
https://beta.cd.screwdriver.cd/pipelines/6512/builds/118224/steps/deploy. It should be set via the environment configuration in sd.yaml

pritamstyz4ever:
Following are the test builds for each scenario of successful, error, abort and timeout cases

https://beta.cd.screwdriver.cd/pipelines/6512/builds/119078/steps/teardown-step1
https://beta.cd.screwdriver.cd/pipelines/6512/builds/119079/steps/teardown-step1
https://beta.cd.screwdriver.cd/pipelines/6512/builds/119077/steps/teardown-step1
https://beta.cd.screwdriver.cd/pipelines/6512/builds/119080/steps/teardown-step1

executor/executor.go
```diff
@@ -320,8 +334,17 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
...
+		exportEnvCommand := []string{
+			"set -e",
+			"export PATH=$PATH:/opt/sd",
+			exportEnvCmd,
+		}
+
 		go func() {
 			runCode, rcErr := doRunCommand(guid, stepFilePath, emitter, f, fReader)
+			// run export file after step execution
+			shargs := strings.Join(exportEnvCommand, " && ")
```
parthasl, line 346: env variable created as part of step is not available by default after each step?

pritamstyz4ever: yeah, it's not, it gets exported as part of the trap that is run after step exit. However, in case of abort that does not happen and so we have to save the export. for normal builds it should be overwritten by the export in trap function.

jithine: it's actually available as part of each step since it's the same shell. 
What we are trying to do is when a new shell i started  (due to step failure or abort or timeout) env variables are exported to it.

parthasl: ok, so existing behavior is teardown steps will not receive any env variable that's set as part of user steps.

parthasl (approved):
👍

parthasl:
@ppaul please add tests.

executor/executor.go
jithine: not exactly, when shell terminates due to step failure, existing `trap` logic works to export env, this new flow is needed only for abort flow, where the `trap` is not getting triggered.

executor/executor.go
```diff
@@ -320,8 +334,17 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
...
+			f.Write([]byte(shargs))
```
jithine, line 347: this could be done inside the case block for `stepAbort` so we can grab the exact state of env variables

pritamstyz4ever: Tested with this change, it doesn't export the env var.
https://beta.cd.screwdriver.cd/pipelines/6512/builds/119164/steps/teardown-step1
https://beta.cd.screwdriver.cd/pipelines/6512/builds/119166/steps/teardown-step1

executor/executor.go
parthasl: to clarify when we abort a build, do sd teardown steps receive the user set environment variables? my assumption is no.

executor/executor.go
parthasl: @pritamstyz4ever checked build timeout doesn't carry the env variables to teardown steps, so this is needed.
only prob I see with f.Write is, it keeps printing this line `set -e && export PATH=$PATH:/opt/sd && tmpfile=/tmp/env_tmp; exportfile=/tmp/` at the end of each step, wondering how its not printing for your build.

executor/executor_test.go
```diff
@@ -475,6 +475,63 @@ func TestTimeout(t *testing.T) {
 	}
 }
 
+func TestTeardownAbort(t *testing.T) {
```
parthasl, line 478: this test is to check if the user teardown step is running correct ? and not abort signal trap.

pritamstyz4ever: added test for notify signal

parthasl (approved):

jithine (approved):

PR #406, opened 2021-03-11
fix: pushgateway url

sugarnaoming:
## Context
There was a difference in the way the `SD_PUSHGATEWAY_URL` protocol was handled in launcher and run.sh, so we will fix it.

## Objective
Add a protocol if `SD_PUSHGATEWAY_URL` does not contain one.

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #407, opened 2021-03-16
fix: Pushgateway url

kkokufud:
## Context
In the [last PR](https://github.com/screwdriver-cd/launcher/pull/406), protocol issues was fixed, but it did not work when port is specified, so we will fix it.

## Objective
- Fixed the problem that it does not work correctly if the pushgateway url includes port. 
- Refactor tests

## License


launch.go
```diff
@@ -72,6 +73,31 @@ func init() {
 	client = retryablehttp.NewClient()
 }
 
+/* has HTTP or HTTPS protocol
+targetUrl => URL
+*/
+func hasHTTPOrHTTPSProtocol(targetUrl string) bool {
```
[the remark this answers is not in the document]
kkokufud: I fixed it.

launch.go
```diff
@@ -72,6 +73,32 @@ func init() {
...
+*/
+func hasHTTPProtocol(targetUrl *url.URL) bool {
+	return strings.Contains(targetUrl.Scheme, "http") || strings.Contains(targetUrl.Scheme, "https")
+}
+
+/* make pushgateway url
+baseUrl => base url for pushgateway
+buildID => sd build id
+*/
+func makePushgatewayUrl(baseUrl string, buildID int) (*url.URL, error) {
```
yuichi10, line 87: From a lint perspective, `Url` should be `URL`.

launch.go
```diff
@@ -72,6 +73,32 @@ func init() {
 	client = retryablehttp.NewClient()
 }
 
+/* has HTTP or HTTPS protocol
+targetUrl => URL
+*/
+func hasHTTPProtocol(targetUrl *url.URL) bool {
+	return strings.Contains(targetUrl.Scheme, "http") || strings.Contains(targetUrl.Scheme, "https")
```
MysticDoll, line 80: golang source code says below. (ref: https://golang.org/src/net/url/url.go?#L428)

> // Maybe rawurl is of the form scheme:path.
> // (Scheme must be [a-zA-Z][a-zA-Z0-9+-.]*)
> // If so, return scheme, path; else return "", rawurl.

Therefore I think we should check only whether `targetUrl.Scheme` is http/https or not.

launch.go
```diff
@@ -72,6 +73,32 @@ func init() {
...
+*/
+func makePushgatewayUrl(baseUrl string, buildID int) (*url.URL, error) {
+	var pushgatewayUrl string = baseUrl
+	u, err := url.Parse(pushgatewayUrl)
+	if err != nil {
+		log.Printf("makePushgatewayUrl: failed to parse url [%v], buildId:[%v], error:[%v]", pushgatewayUrl, buildID, err)
+		return nil, err
+	}
+
+	if !hasHTTPProtocol(u) {
```
yuichi10, line 95: Is it better to set http or https, if there is no Schema?

launch.go
```diff
@@ -83,15 +110,7 @@ func pushMetrics(status string, buildID int) error {
...
-		url, err := url.Parse(os.Getenv("SD_PUSHGATEWAY_URL"))
-		if err != nil {
-			log.Printf("pushMetrics: failed to parse url [%v], buildId:[%v], error:[%v]", os.Getenv("SD_PUSHGATEWAY_URL"), buildID, err)
-			return nil
-		}
-		if url.Scheme == "" {
-			url.Scheme = "http"
-		}
-		url.Path = url.Path + "/metrics/job/containerd/instance/" + strconv.Itoa(buildID)
+		pushgatewayUrl, err := makePushgatewayUrl(os.Getenv("SD_PUSHGATEWAY_URL"), buildID)
```
yuichi10, line 113: Do you need to check to err?

launch.go
```diff
@@ -72,6 +73,32 @@ func init() {
...
+	u, err := url.Parse(pushgatewayURL)
+	if err != nil {
+		log.Printf("makePushgatewayURL: failed to parse url [%v], buildId:[%v], error:[%v]", pushgatewayURL, buildID, err)
+		return nil, err
+	}
+
+	if !hasHTTPProtocol(u) {
+		return nil, errors.New("Pushgateway url has no http/https protocol. Please make sure it.")
+	}
+	u.Path = u.Path + "/metrics/job/containerd/instance/" + strconv.Itoa(buildID)
```
yuichi10, line 98: Is it better to use path.Join?

If baseURL is like 'http://sample.com/', u.Path will be 'http://sample.com//metrics/job/containerd/instance/id'

data/emitter
```diff
@@ -1,9 +0,0 @@
-{"t":1608079158474,"m":"Screwdriver Launcher information","s":"sd-setup-launcher"}
```
yuichi10, line 1: launch_test looks like to use this file. 
Can you delete this file?
https://github.com/screwdriver-cd/launcher/blob/d99488f7fe3723ffb4d29d345bf4de413b9bbbf0/launch_test.go#L27

launch.go
kkokufud: I fixed.

launch.go
kkokufud: I fixed.

launch.go
kkokufud: I fixed.

data/emitter
kkokufud: `./data/emitter` is created automatically, so we don't need it.
It looks like it was [deleted](https://github.com/screwdriver-cd/launcher/commit/24c4e3f4d7787113d831c096b88ac7e22358e08a#diff-a680012bcc3aec2d5edf07c5b0b31bbccef0bc5384d733ba73af2e54c9909dea) in the past, but it was restored. It's also listed in [.gitignore](https://github.com/screwdriver-cd/launcher/blob/d99488f7fe3723ffb4d29d345bf4de413b9bbbf0/.gitignore#L5).

launch.go
kkokufud: I fixed it.

launch.go
kkokufud: I fixed it.

MysticDoll (approved):

launch.go
```diff
@@ -72,6 +73,31 @@ func init() {
...
+func makePushgatewayURL(baseURL string, buildID int) (string, error) {
+	var pushgatewayURL string = baseURL
+	u, err := url.Parse(pushgatewayURL)
+	if err != nil {
+		log.Printf("makePushgatewayURL: failed to parse url [%v], buildId:[%v], error:[%v]", pushgatewayURL, buildID, err)
+		return "", err
+	}
+	u.Path = path.Join(u.Path, "/metrics/job/containerd/instance/"+strconv.Itoa(buildID))
+	if !hasHTTPProtocol(u) {
+		return "http://" + u.String(), nil
```
yuichi10, line 96: How about this?
```suggestion
		u.Scheme = "http"
```

yuichi10: Can you add a test that adds a schema when an environment variable with no schema is specified?

kkokufud: If pushgateway url has port and no protocol, `url.Parse` behaves strangely. So we can't use `u.Scheme` in this case.
Trying code is below. I fixed it for this case.
```
package main
import (
	"fmt"
	"log"
	"net/url"
)
func main() {
	u, err := url.Parse("pushgateway.monitoring:9091")
	if err != nil {
		log.Fatal(err)
	}
	if u.Scheme == "" {
		u.Scheme = "http"
	}
	fmt.Println(u.Scheme)  // => pushgateway.monitoring
	u.Path = u.Path + "/metrics/job/containerd/instance/"
	fmt.Println(u.String()) // => pushgateway.monitoring:9091
}
```

kkokufud: > Can you add a test that adds a schema when an environment variable with no schema is specified?

I added a test when `SD_BUSHGATEWAY_URL` has no scheme.

yuichi10 (approved):
LGTM

tkyi (approved):

PR #410, opened 2021-03-18
fix: Always merge event meta first

tkyi:
## Context

Currently, event meta is only set if none of the other cases are true (join, one parent build, parent event). This can cause problems when the user wants to get keys from event meta as well.

## Objective

This PR always merged event meta first, then handles other cases.

## References

N/A

## License


parthasl (approved):
👍

parthasl:
@tkyi no change in tests?

PR #404, opened 2021-03-01
fix: Env variables not available in teardown steps

parthasl:
## Context

Env variables not available in teardown steps

## Objective

Capturing Interrupt and Term signal works differently for different shells. Update trap to handle ABRT signal to capture signal as expected.

## References

https://github.com/screwdriver-cd/launcher/pull/401
https://github.com/screwdriver-cd/launcher/pull/396
https://github.com/screwdriver-cd/launcher/pull/394
https://github.com/screwdriver-cd/launcher/pull/388

Related to https://github.com/screwdriver-cd/screwdriver/issues/2361

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #413, opened 2021-04-15
fix(2351): teardown steps do not get the environment variables

parthasl:
## Context

Environment variables are not available in teardown steps.

## Objective

Abort or Timeout doesn't capture the environment variable in teardown steps when `sleep` command is used in the sd step.  `sleep` is not killed via the SIGABRT signal and keeps running in the background. Explicitly kill `sleep` command for abort (or) timeout signal and after all teardown steps are executed.

## References

https://github.com/screwdriver-cd/screwdriver/issues/2351
https://github.com/screwdriver-cd/screwdriver/issues/2361
https://github.com/screwdriver-cd/launcher/pull/404

## License

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -2,6 +2,7 @@ package executor
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -354,6 +355,8 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
 				code = 3
 			}
 			_ = c.Process.Signal(syscall.SIGABRT)
+			terminateSleep(shellBin, sourceDir, true) // kill all running sleep
+
 		case stepAbort := <-sig:
 			f.Write([]byte{4})
 			if firstError == nil {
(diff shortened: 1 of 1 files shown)


pritamstyz4ever (approved):

jithine (approved):

executor/executor.go
```diff
@@ -396,6 +400,25 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
...
+	if killAll {
+		cmdStr = "pids=$(ps -ef | grep '[s]leep' | awk '{print $2}'); if [ ! -z $pids ]; then kill $pids; else echo $pids; fi;"
+	}
+	shargs = append(shargs, cmdStr)
+	c := exec.Command(shellBin, shargs...)
+	c.Stdout = &stdout
+	c.Stderr = &stderr
+	c.Dir = sourceDir
+	err := c.Run()
+	if err != nil || strings.TrimSpace(stderr.String()) != "" {
```
jithine, line 421: does this makes sure that this function never errors out and worst case it won't kill sleep, but continue with rest of execution ?

parthasl: yes, it will continue with the rest of execution even if it errors.

PR #421, opened 2021-07-12
[2] fix(2501): set build parameter explicitly

yoshwata:
## Context

fix https://github.com/screwdriver-cd/screwdriver/issues/2501

## Objective

The parameters of the parent event overwrote the parameters of the current build during event restart. By explicitly setting the build parameters, the parameters set by the user will be used preferentially.

## References

Please merge after docker build become stable
https://github.com/screwdriver-cd/launcher/pull/420

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #422, opened 2021-07-14
fix(2501): Fetch parent meta.build if the key is not reserved

kumada626:
## Context

meta data set on `build` key is now lost because it's overwriting build key.
https://github.com/screwdriver-cd/launcher/pull/421#discussion_r669215113
We should set properly `build` key meta.

## Objective

I fixed to deepMerge parent `build` meta and default `build` meta.
If there are the same keys, it should use default `build` meta.
Otherwise, it should set parent `build` meta.

## References

https://github.com/screwdriver-cd/screwdriver/issues/2501

## License

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -511,7 +511,11 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 		buildMeta["coverageKey"] = coverageInfo.EnvVars["SD_SONAR_PROJECT_KEY"]
 	}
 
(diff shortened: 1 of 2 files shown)


yoshwata (approved):

wahapo (approved):

jithine:
Verified https://cd.screwdriver.cd/pipelines/7217/builds/721856 Thank you for the quick fix 🙇

PR #429, opened 2021-11-11
feat(#1658): Add flag to skip habitat installation

pritamstyz4ever:
## Context

We need to have a flag to allow skipping habitat installation for AWS Integration setup

## Objective

This PR adds a flag to skip habitat processing in the launcher and update `launch` command for AWS integration compatibility

## References

https://github.com/screwdriver-cd/screwdriver/issues/1658

## License


Docker/launcher_entrypoint.sh
```diff
@@ -52,26 +52,29 @@ smart_run whoami
...
-
-# Create directory for hab pkg binlink destination
-smart_run mkdir -p /usr/sd/bin
-
-# Binlinking bash from core/bash into /bin
-if ! binary_exists bash; then
-    smart_run /opt/sd/bin/hab pkg binlink core/bash bash || echo 'Failed to symlink bash'
-fi
+
+if ! $SD_NO_HAB; then
```
jithine, line 56: can we call it `SD_HAB_ENABLED` and this check shouild change to `if [ "$SD_HAB_ENABLED" = "yes" ];`

pritamstyz4ever: ok, doing the inverse means, we need to change code in all executors again to make it work

jithine: I feel it's more readable with `SD_HAB_ENABLED`, it took me a while to write the if logic . with inverse. let's update executors.

pritamstyz4ever: Updated the flag name!

tkyi: can we set to `true` vs `yes` just to be consistent?

jithine (approved):

PR #431, opened 2021-12-16
Revert "fix: try to use bash shell for launcher"

jithine:
Reverts screwdriver-cd/launcher#430


adong (approved):

jithine:
Looks like we made a release with this version internally and it's working. Now I am not sure why we got the error in cd.screwdriver.cd 🤔  cc @pritamstyz4ever

PR #438, opened 2022-04-28
feat: Make launcher docker file multi-arch compatible

pritamstyz4ever:
## Context

Screwdriver needs to support arm64 containers for building

## Objective

This PR updates dockerfile to be multi-arch compatible

## License

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -6,6 +6,7 @@ builds:
       - darwin
     goarch:
       - amd64
+      - arm64
     # Include the default settings from https://goreleaser.com/#builds
     # Also include static compilation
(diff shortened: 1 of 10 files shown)


DekusDenial (approved):
i can approve this for now, but u need to be aware that there are some binaries here aren’t compatible with arm yet and build dependency on hab is still a pending problem, so there should other follow up to address them

pritamstyz4ever:
> i can approve this for now, but u need to be aware that there are some binaries here aren’t compatible with arm yet and build dependency on hab is still a pending problem, so there should other follow up to address them

I agree this will need revision based on version updates and new binaries. Thanks

PR #439, opened 2022-06-15
feat: add SD_STEP_NAME env variable

jacobtolar:
## Context

As a user, I want to access the current step name from code executing within the step. 

## Objective

Add new environment variable SD_STEP_NAME set before each ordinary step is executed run. Only set for ordinary user steps, not setup or teardown commands (during teardown, SD_STEP_NAME would be set to the last successfully executed step). If desired I could change this behavior.

## License

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -138,9 +138,11 @@ func doRunSetupCommand(emitter screwdriver.Emitter, f *os.File, r io.Reader, set
 	return nil
 }
 
-func doRunCommand(guid, path string, emitter screwdriver.Emitter, f *os.File, fReader io.Reader) (int, error) {
+func doRunCommand(guid, path string, emitter screwdriver.Emitter, f *os.File, fReader io.Reader, stepName string) (int, error) {
 	executionCommand := []string{
 		"export SD_STEP_ID=" + guid,
+		// escape not necessary because step name is limited to [A-Za-z0-9_-]
+		";export SD_STEP_NAME=" + stepName,
 		";. " + path,
 		";echo",
 		";echo " + guid + " $?\n",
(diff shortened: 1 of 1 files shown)


executor/executor.go
```diff
@@ -138,9 +138,11 @@ func doRunSetupCommand(emitter screwdriver.Emitter, f *os.File, r io.Reader, set
...
 }
 
-func doRunCommand(guid, path string, emitter screwdriver.Emitter, f *os.File, fReader io.Reader) (int, error) {
+func doRunCommand(guid, path string, emitter screwdriver.Emitter, f *os.File, fReader io.Reader, stepName string) (int, error) {
 	executionCommand := []string{
 		"export SD_STEP_ID=" + guid,
+		// escape not necessary because step name is limited to [A-Za-z0-9_-]
+		";export SD_STEP_NAME=" + stepName,
 		";. " + path,
 		";echo",
```
jacobtolar, line 147: Not related to my change, but I noticed that the commit introducing this 'echo' statement (2556c791 / #109) unintentionally broke the status check on the next line (`$?` comes from `echo` rather than from `source`).

It shouldn't really make a difference (i.e. `echo guid $?` would be skipped because of `set -e` if the user code results in anything other than success). But it's not the original intent of the code (and confusing). It should either just be `echo 0` (because we know the line will only ever be reached on success) or something like `"; printf \\\\n " + guid + " $?\n"` and the extraneous `echo` removed.

If you have a preferred fix among those options, I am happy to send a separate pr.

executor/executor.go
```diff
@@ -138,9 +138,11 @@ func doRunSetupCommand(emitter screwdriver.Emitter, f *os.File, r io.Reader, set
 	return nil
 }
 
-func doRunCommand(guid, path string, emitter screwdriver.Emitter, f *os.File, fReader io.Reader) (int, error) {
+func doRunCommand(guid, path string, emitter screwdriver.Emitter, f *os.File, fReader io.Reader, stepName string) (int, error) {
 	executionCommand := []string{
 		"export SD_STEP_ID=" + guid,
+		// escape not necessary because step name is limited to [A-Za-z0-9_-]
```
jacobtolar, line 144: This is described here: https://github.com/screwdriver-cd/data-schema/blob/b7cf7ab0bc5039570aeb03daa790f0d67e183544/config/job.js#L92-L103

executor/executor.go
DekusDenial: sure, u can use that printf or echo, whichever can wrap a newline before and after will do

tkyi (approved):

PR #440, opened 2022-09-02
[WIP] fix(2715): Some process do not stop after build timed out

ibu1224:
## Context

Fix the following Issue
https://github.com/screwdriver-cd/screwdriver/issues/2715

## Objective

This problem is caused by SIGABRT unable to stop processes created other than  described in sd-step.
In other words, if a grandchild process is created, the process does not stop and continues to run.

As a solution, modified the system to send a SIGTERM to the created process group.
Sending kill to `-PID` will send a signal to that process group.
https://man7.org/linux/man-pages/man2/kill.2.html

If this solution cannot be adopted, it is better to accept it as a specification in Screwdriver.cd.

Well-behaved programs create/process child processes themselves.
If the user defines the sd-step, there is also the idea that the user is responsible for it.

## References

- https://github.com/screwdriver-cd/screwdriver/issues/2715
- https://man7.org/linux/man-pages/man2/kill.2.html
## License

https://github.com/screwdriver-cd/screwdriver/issues/2715
https://man7.org/linux/man-pages/man2/kill.2.html

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -356,7 +356,7 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
 				firstError = buildTimeout
 				code = 3
 			}
-			_ = c.Process.Signal(syscall.SIGABRT)
+			syscall.Kill(-c.Process.Pid, syscall.SIGKILL)
 			terminateSleep(shellBin, sourceDir, true) // kill all running sleep
 
 		case stepAbort := <-sig:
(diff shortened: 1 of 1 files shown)


y-oksaku (approved):

itleigns (approved):

kumada626 (approved):

tkyi (approved):

jithine:
is this tested in kubernetes context ? any surprises we need to be aware of due to this change ? Does it cause any interruption to signal handling  https://github.com/screwdriver-cd/launcher/blob/master/Docker/run.sh#L4

ibu1224:
@jithine 
As we talked about in slack, there was a flaw in our implementation, so the PR is CLOSED.
I will comment on the issue for more details.

PR #452, opened 2023-05-03
fix: add SHELLOPTS to ignore list in teardown step

adong:
## Context

To address issues like
```
/tmp/env_export: line 110: SHELLOPTS: readonly variable
```

When user has script in sd.yaml such as

```
      steps:
        - env: |
            set -x; # export SHELLOPTS
            env
```

We will have issue like: 

image

## Objective
This PR add `SHELLOPTS` to be ignored

Test pending

## License

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -276,7 +276,7 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
 	// Command to Export Env. Use tmpfile just in case export -p takes some time
 	exportEnvCmd :=
 		"tmpfile=" + tmpFile + "; exportfile=" + exportFile + "; " +
-			"export -p | grep -vi \"PS1=\" > $tmpfile && mv -f $tmpfile $exportfile; "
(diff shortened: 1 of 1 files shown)


executor/executor.go
```diff
@@ -276,7 +276,7 @@ func Run(path string, env []string, emitter screwdriver.Emitter, build screwdriv
 	// Command to Export Env. Use tmpfile just in case export -p takes some time
 	exportEnvCmd :=
 		"tmpfile=" + tmpFile + "; exportfile=" + exportFile + "; " +
-			"export -p | grep -vi \"PS1=\" > $tmpfile && mv -f $tmpfile $exportfile; "
+			"export -p | grep -vi \"PS1=|SHELLOPTS=\" > $tmpfile && mv -f $tmpfile $exportfile; "
```
jithine, line 279: Has this bees tested against different shells ? We should do a more extensible approach to add more variables in future

jithine (changes requested):

PR #454, opened 2023-06-30
fix(2878): exclude warning from merging build meta

klu909:
## Context

build warning should be specific to the build
## Objective

exclude build warning from inheriting to downstream builds
## License


tkyi (approved):
Please update description

PR #456, opened 2023-07-28
feat: Add source directory to Sonarqube project name

yk634:
## Context

If there are multiple projects in one repository, the source directory can be set for the pipeline, but the source directory is not set for the project name in SonarQube.

In SonarQube UI, it is not possible to distinguish which directory the project was created from, so modify the source directory so that it is set to the project name.

## Objective

In pipelines where the source directory is set, modify it so that it is set in the format `repositoryName:sourceDir:jobName`.

image

## License

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -442,11 +442,21 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 	oldJobName := job.Name
 	pr := prNumber(job.Name)
 
+	scm, err := parseScmURI(pipeline.ScmURI, pipeline.ScmRepo.Name)
+	if err != nil {
+		return err, "", ""
+	}
+
+	pipelineName := pipeline.ScmRepo.Name
+	if scm.RootDir != "" {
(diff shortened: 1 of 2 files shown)


launch.go
```diff
@@ -442,11 +442,21 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
...
 	pr := prNumber(job.Name)
 
+	scm, err := parseScmURI(pipeline.ScmURI, pipeline.ScmRepo.Name)
+	if err != nil {
+		return err, "", ""
+	}
+
+	pipelineName := pipeline.ScmRepo.Name
+	if scm.RootDir != "" {
+		pipelineName = pipeline.ScmRepo.Name + ":" + scm.RootDir
```
y-oksaku, line 452: How about using `+=` operator?

yk634: fixed

y-oksaku (approved):

tkyi (approved):

PR #464, opened 2023-10-29
fix: add /usr/sd-static/bin to PATH

DekusDenial:
provide bin path to statically linked executables

## License

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -157,7 +157,7 @@ func doRunCommand(guid, path string, emitter screwdriver.Emitter, f *os.File, fR
 // Executes teardown commands
 func doRunTeardownCommand(cmd screwdriver.CommandDef, emitter screwdriver.Emitter, shellBin, exportFile, sourceDir string, stepExitCode int) (int, error) {
 	shargs := []string{"-e", "-c"}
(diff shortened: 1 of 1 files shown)


klu909 (approved):

executor/executor.go
```diff
@@ -157,7 +157,7 @@ func doRunCommand(guid, path string, emitter screwdriver.Emitter, f *os.File, fR
 // Executes teardown commands
 func doRunTeardownCommand(cmd screwdriver.CommandDef, emitter screwdriver.Emitter, shellBin, exportFile, sourceDir string, stepExitCode int) (int, error) {
 	shargs := []string{"-e", "-c"}
-	cmdStr := "export PATH=${PATH}:/opt/sd:/usr/sd/bin SD_STEP_EXIT_CODE=" + strconv.Itoa(stepExitCode) + " && " +
+	cmdStr := "export PATH=${PATH}:/opt/sd:/usr/sd/bin:/usr/sd-static/bin SD_STEP_EXIT_CODE=" + strconv.Itoa(stepExitCode) + " && " +
```
jithine, line 160: why not reuse existing `/usr/sd/bin` ?

DekusDenial: like what the name implies, need to separate static and dynamic binaries

adong:
I was told the use case is to use statically compiled binaries
@kumada626

DekusDenial:
like git, as we tried to move away from using habitat

PR #466, opened 2023-11-09
fix: ensure right event dumping

DekusDenial:
## License

### DIFF
--- a/Docker/run.sh
+++ b/Docker/run.sh
@@ -3,10 +3,12 @@
 # Trap these SIGNALs and run teardown
 trap 'trap_handler $@' INT TERM
 
+event_dumped=""
 dump_event() {
-  if [ -r /tmp/sd_event.json ]; then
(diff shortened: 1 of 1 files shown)


adong (approved):

jithine (approved):

Docker/run.sh
```diff
@@ -3,16 +3,20 @@
 # Trap these SIGNALs and run teardown
 trap 'trap_handler $@' INT TERM
 
+event_dumped=""
 dump_event() {
-  if [ -r /tmp/sd_event.json ]; then
+  if [ -r /tmp/sd_event.json ] && [ -z "$event_dumped" ]; then
```
adong, line 8: Question, does `event_dumped` set elsewhere? Otherwise, how does this ever get called?

DekusDenial: it's set up above

adong: Nevermind, it means `dump_event` only gets called 1 time

PR #468, opened 2023-11-28
fix(2952): encode percent of rootDir

yk634:
## Context

This PR is for https://github.com/screwdriver-cd/screwdriver/issues/2952#issuecomment-1809697640.

Pipeline build with rootDir containing `%` fails at `sd-setup-launcher` step.

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
```

## Objective

`url.Parse()` fails at the point where it encodes `%`, so replace `%` with `%25` before that.

## License

### DIFF
--- a/screwdriver/screwdriver.go
+++ b/screwdriver/screwdriver.go
@@ -338,7 +338,9 @@ func (a api) GetAPIURL() (string, error) {
 
 // Get coverage object with coverage information
 func (a api) GetCoverageInfo(jobID, pipelineID int, jobName, pipelineName, scope, prNum, prParentJobId string) (coverage Coverage, err error) {
-	url, err := a.makeURL(fmt.Sprintf(CoverageURL, jobID, pipelineID, jobName, pipelineName, scope, prNum, prParentJobId))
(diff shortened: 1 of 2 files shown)


screwdriver/screwdriver.go
```diff
@@ -338,7 +338,9 @@ func (a api) GetAPIURL() (string, error) {
 
 // Get coverage object with coverage information
 func (a api) GetCoverageInfo(jobID, pipelineID int, jobName, pipelineName, scope, prNum, prParentJobId string) (coverage Coverage, err error) {
-	url, err := a.makeURL(fmt.Sprintf(CoverageURL, jobID, pipelineID, jobName, pipelineName, scope, prNum, prParentJobId))
+	// If pipelineName contains a percent sign (%), it needs to be encoded to ensure the URL can be parsed correctly.
+	encodedPipelineName := strings.Replace(pipelineName, "%", "%25", -1)
```
tkyi, line 342: should we make constants for any of these?

tkyi (approved):
Small comment

screwdriver/screwdriver.go
yk634: I think the constant is unnecessary because characters other than "percent" do not need to be replaced.

PR #478, opened 2024-11-05
feat(3234): event and parentEvent meta should be mutually exclusive

VonnyJap:
## Context

The current implementation of meta behavior causes the merging of parentEvent and event altogether whenever they are available. 

## Objective

This implementation should not be necessary, instead event and parentEvent meta should be mutually exclusive. The parentEvent will only be relevant for the initial build of the restart. Afterward, it will simply be part of the event, as the first build will update the event appropriately during the launcher exit.

## References

https://github.com/screwdriver-cd/screwdriver/issues/3234

## License

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -476,7 +476,7 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 
 	// Note: event and parent event meta are mutually exclusive
 	// the first event in the build chain for a restart case will use the parent event meta
-	// on launcher exit, the meta will be updated to the current event meta
+	// on launcher exit, the build meta will be merged int to the current event meta
 
 	// merge event meta if available
 	if len(event.Meta) > 0 {
--- a/launch.go
+++ b/launch.go
@@ -475,7 +475,7 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 	}
 
 	// Note: event and parent event meta are mutually exclusive
-	// the first event in the build chain for a restart case will use the parent event meta
+	// the first build in the event for a restart case will use the parent event meta
 	// on launcher exit, the build meta will be merged int to the current event meta
 
 	// merge event meta if available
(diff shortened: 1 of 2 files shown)


launch.go
```diff
@@ -474,8 +474,18 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
...
+	// on launcher exit, the meta will be updated to the current event meta
+
+	// merge event meta if available
+	if len(event.Meta) > 0 { // If has meta, marshal it
+		log.Printf("Fetching Event Meta JSON %v", event.ID)
+		if event.Meta != nil {
+			mergedMeta = deepMergeJSON(mergedMeta, event.Meta)
+		}
+	} else if event.ParentEventID != 0 { // otherwise, fetch from parent event
+		// If has parent event, fetch meta from parent event
```
sagar1312, line 488: I believe the code is self-explanatory
```suggestion
	if len(event.Meta) > 0 {
		log.Printf("Fetching Event Meta JSON %v", event.ID)
		if event.Meta != nil {
			mergedMeta = deepMergeJSON(mergedMeta, event.Meta)
		}
	} else if event.ParentEventID != 0 {
```

launch.go
```diff
@@ -474,8 +474,18 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
 		return err, "", ""
 	}
 
-	// Always merge parent event meta if it exists (Issue #3234)
-	if event.ParentEventID != 0 { // If has parent event, fetch meta from parent event
+	// Note: event and parent event meta are mutually exclusive
+	// the first event in the build chain for a restart case will use the parent event meta
```
sagar1312, line 478:
```suggestion
	// the first build in the event for a restart case will use the parent event meta
```

launch.go
```diff
@@ -474,8 +474,18 @@ func launch(api screwdriver.API, buildID int, rootDir, emitterPath, metaSpace, s
...
+	// on launcher exit, the meta will be updated to the current event meta
```
sagar1312, line 479:
```suggestion
	// on launcher exit, the build meta will be merged int to the current event meta
```

tkyi (approved):

sagar1312 (approved):

